PR #32, opened 2022-03-07
Source code and tests for Playlist and Song data structures

nathanduchesne:
If applied, this commit will add the possibility of creating playlists, which contain songs. Songs are equal if their artist name and title are identical, because we can safely assume an artist doesn't have two songs with the same name. Playlists can be added to one another.


nathanduchesne:
Works now!

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
```diff
@@ -0,0 +1,44 @@
+class Playlist (playListName: String, songSet: Set<Song>, genreType: Genre){
+    var name: String = playListName
+    private val songs: MutableSet<Song>
+    var genre: Genre = genreType
+
+    init {
+        val result = mutableSetOf<Song>()
+        for (song in songSet){
```
VictorSchneuwly, line 8: Why don't you use '.addAll(songSet)' here ?

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    fun removeSongs(removedSongs: Set<Song>){
+        songs.removeAll(removedSongs)
+    }
+
+    fun combinePlaylists(other: Playlist, newName: String = name, newGenre: Genre = Genre.NONE) {
+        name = newName
+        genre = newGenre
+        for (song in other.songs) {
```
VictorSchneuwly, line 36: Same here, why not use 'addAll' ?

app/src/main/java/com/github/fribourgsdp/radio/Song.kt
```diff
@@ -0,0 +1,38 @@
+import java.util.*
+
+class Song (songName: String, artistName: String, songLyrics: String) {
+
+    var name: String = reformat_song_name(songName)
+        get() {return field}
+        set(value) {field = reformat_song_name(value)}
+
+    var artist: String = artistName
+    var lyrics: String = songLyrics
```
VictorSchneuwly, line 10: Why not make all of these as `val` to make a `Song` immutable?

nathanduchesne: For custom playlists made by users I thought it would be nice to allow users the chance to change the song name.

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
nathanduchesne: Same as above, I did this to avoid the fact that if the owner of the 'other' playlist changes a song name, it doesn't change in our new playlist that we combined previously.

app/src/main/java/com/github/fribourgsdp/radio/Song.kt
```diff
@@ -0,0 +1,38 @@
+import java.util.*
+
+class Song (songName: String, artistName: String, songLyrics: String) {
+
+    var name: String = reformat_song_name(songName)
+        get() {return field}
+        set(value) {field = reformat_song_name(value)}
+
+    var artist: String = artistName
```
Flamantrose321, line 9: Do you not reformat the artist name?

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
```diff
@@ -0,0 +1,46 @@
...
+            songs.add(Song(song.name, song.artist, song.lyrics))
+        }
+    }
+
+    fun getSongs(): MutableSet<Song> {
+        return songs.toMutableSet()
```
shanzhaiii, line 44: Would it not be better to only return an immutable copy of the playlist?

build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath "com.android.tools.build:gradle:7.0.3"
```
shanzhaiii, line 8: Why was this changed?

PR #35, opened 2022-03-09
Song and Playlist data structures

nathanduchesne:
If applied, this commit will add the possibility of creating playlists, which contain songs. Songs are equal if their artist name and title are identical, because we can safely assume an artist doesn't have two songs with the same name. Playlists can be added to one another.

Note: I had to create a new branch due to codeclimate problems on the other branch and an incompatible gradlew file.


sim0mo (approved):
Good job. An idea for a future task : flexibility in song comparisons (e.g. commas, quotes etc.) But this is another story.

shanzhaiii (approved):
Perfect. I think using toset is much better

PR #36, opened 2022-03-10
User Class and User Profile Activity

shanzhaiii:


app/src/main/java/com/github/fribourgsdp/radio/User.kt
```diff
@@ -0,0 +1,52 @@
...
+    val color : Int = Color.rgb(
+        Random.nextInt(100, 200),
+        Random.nextInt(100, 200),
+        Random.nextInt(100, 200))
+    private var linkedSpotify: Boolean = false
+    val initial get(): Char = name.elementAt(0)
+    val spotifyLinked get(): Boolean = linkedSpotify
+
+
+    private fun nameSanitizer(name: String) {
```
BartLucien, line 27: I think the sanitizer will need to be more robust in the future

app/src/main/java/com/github/fribourgsdp/radio/User.kt
```diff
@@ -0,0 +1,52 @@
...
+        }
+    init {
+        nameSanitizer(name)
+    }
+    private val playlists = mutableSetOf<Playlist>()
+    val color : Int = Color.rgb(
+        Random.nextInt(100, 200),
+        Random.nextInt(100, 200),
+        Random.nextInt(100, 200))
+    private var linkedSpotify: Boolean = false
```
sim0mo, line 22: Isn't it a bit early to add this variable regarding spotify since it is not yet integrated ? Remember "code for what you need, not what you may need in the future"

app/src/main/res/layout/activity_user_profile.xml
```diff
@@ -0,0 +1,88 @@
+<?xml version="1.0" encoding="utf-8"?>
```
BartLucien, line 1: I think we should do what cirrus requests.

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
```diff
@@ -4,6 +4,9 @@ import android.content.Intent
 import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
 import android.widget.Button
+import android.widget.ImageButton
+
+const val USERNAME = "com.github.fribourgsdp.radio.USERNAME"
```
VictorSchneuwly, line 9: You shouldn't really put that simply in the file. You should put it in a companion object of the activity you want to send it to. Then you call `ActivityName.USERNAME` here.

VictorSchneuwly (dismissed):

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
shanzhaiii: Agreed, this is a temporary solution to use intents to show that the UI is dynamic. In the future, we will use preferences, but we did not have time to implement it in this sprint

app/src/main/java/com/github/fribourgsdp/radio/User.kt
shanzhaiii: Yes you are correct, this will definitely change. This was just a way to test for a dynamic UI

Flamantrose321 (approved):

sim0mo (approved):

PR #59, opened 2022-03-12
Google sign in integration

Flamantrose321:
Added googleSignIn activity with tests


app/build.gradle
```diff
@@ -33,21 +34,37 @@ android {
...
 
-dependencies {
+apply plugin: 'com.google.gms.google-services'
 
+dependencies {
     implementation 'androidx.core:core-ktx:1.7.0'
     implementation 'androidx.appcompat:appcompat:1.4.1'
     implementation 'com.google.android.material:material:1.5.0'
     implementation 'androidx.constraintlayout:constraintlayout:2.1.3'
+
```
BartLucien, line 52: Does it also work with the new version?

app/src/main/java/com/github/fribourgsdp/radio/GoogleSignInActivity.kt
```diff
@@ -0,0 +1,140 @@
...
+
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        binding = ActivityGoogleSignInBinding.inflate(layoutInflater)
+        setContentView(binding.root)
+
+        //configure for the Google SignTn
+        val googleSignInOptions = GoogleSignInOptions.Builder(GoogleSignInOptions.DEFAULT_SIGN_IN)
+            .requestIdToken(getString(R.string.default_web_client_id))
```
VictorSchneuwly, line 46: This reference is not found in `string.xml`. Am I missing something ?

app/build.gradle
Flamantrose321: yes it's work too, good remark

VictorSchneuwly:
Also, when I try to log in, it always says "cancelled".

app/src/main/java/com/github/fribourgsdp/radio/GoogleSignInActivity.kt
Flamantrose321: The reference is in a values.xml file generated automatically during the build under app/build/generated/res/google-services/release/values but the android studio interface doesn't always find it. But when you build the app, this is never a problem and the reference is found.

sim0mo:
> > Also, when I try to log in, it always says "cancelled".
> 
> I have actually seen this bug too, it seems that before every successful Google login there is a fail before. However, this does not affect the functionality of the program. So I will fix this later.

It never works for me as well... Even if I try multiple times

nathanduchesne:
> > > Also, when I try to log in, it always says "cancelled".
> > 
> > 
> > I have actually seen this bug too, it seems that before every successful Google login there is a fail before. However, this does not affect the functionality of the program. So I will fix this later.
> 
> It never works for me as well... Even if I try multiple times

Same here.. It seems to work since after signing in once it remembers my profile but if I click on it, it says Cancelled. Are we sure the error message is correct?

nathanduchesne (approved):
After various tests of the different members after pulling this branch locally, the feature now works.

PR #61, opened 2022-03-15
Game data structure

VictorSchneuwly:
Create the basic Game data structure.


app/src/main/java/com/github/fribourgsdp/radio/Game.kt
```diff
@@ -0,0 +1,117 @@
...
+        val user = listUser[0]
+        Collections.rotate(listUser, 1)
+        if (usersToPlay == 0) {
+            ++currentRound
+            usersToPlay = listUser.size
+        }
+        --usersToPlay
+
+        return user
+    }
```
nathanduchesne, line 43: When will .currentRound be accessed? This implementation works but means that for a short time period, we fetch a user to play for round i and have currentRound set to i - 1.

ex. 2 players, 2 rounds:
getUserToPlay() returns Player1, then decrements usersToPlay, getUserToPlay() returns Player2, decrements usersToPlay. When we call getUserToPlay() again, this means we will be in round 2 since both players have already played once. However, currentRound will still be 1. 

Obviously this isn't a problem as long as we know this is how the method is implemented, but it could be problematic when calling this function later on.

nathanduchesne (approved):

app/src/main/java/com/github/fribourgsdp/radio/Game.kt
VictorSchneuwly: `currentRound` is init to 1 and not 0, so here it would be round 2, which is correct.

nathanduchesne: No what I mean is that currentRound is only updated once we start choosing the first player for this new round. If you look at my example, you'll understand that after the 2 first players are chosen, we would imagine currentRound should be 2. However, it only updates itself to 2 once we call getUserToPlay() a third time. This means that when we call getUserToPlay() for the start of round 2, our game still considers we are in the first round, which I find slightly confusing.

VictorSchneuwly: Ok, I see what you mean now. In my mind, a round starts at the same time that a user is picked, hence the update here.

BartLucien (approved):
It looks good to me

PR #62, opened 2022-03-16
Musixmatch getter

sim0mo:


app/src/main/java/com/github/fribourgsdp/radio/LyricsGetter.kt
```diff
@@ -102,6 +127,28 @@ class LyricsGetter {
             }
         }
 
+        private fun cleanLyrics(lyrics : String) : String{
```
BartLucien, line 130: I think you should document why you want to suppress the last 4 lines, even right now, it is clear why.

app/src/main/java/com/github/fribourgsdp/radio/Song.kt
```diff
@@ -1,17 +1,24 @@
 package com.github.fribourgsdp.radio
 
 import java.util.*
+import java.util.concurrent.CompletableFuture
 
-class Song (songName: String, artistName: String, songLyrics: String) {
+class Song (songName: String, artistName: String, private val songLyrics: CompletableFuture<String>) {
```
BartLucien, line 6: Wouldn't be a way to force the completion of the songLyrics in the constructor so that we don't have to play around with a CompletableFuture? Like doing what to do in line 15 in the constructor directly?
edit: typo

app/src/main/java/com/github/fribourgsdp/radio/LyricsGetter.kt
```diff
@@ -5,37 +5,62 @@ import org.json.JSONException
...
             try {
                 trackID = trackIDFuture.get()
-            } catch (e : Throwable){
+            } catch (e: Throwable) {
                 future.complete(LYRICS_NOT_FOUND)
                 return future
             }
-            val url = BASE_URL + "track.lyrics.get?track_id=" + trackID + "&apikey=" + API_KEY
+            val url =
+                BASE_URL + "track.lyrics.get?track_id=" + trackID.toString() + "&apikey=" + API_KEY
```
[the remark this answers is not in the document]
sim0mo: Ok, why not

app/src/main/java/com/github/fribourgsdp/radio/Song.kt
sim0mo: I thought doing it this way would reduce the waiting time, delaying the get() to the point where we really need it. We don't "play around", given that it is private and that when we get this property, a call to get() is issued.

BartLucien (approved):
It is good for me

VictorSchneuwly (approved):
Good work!

PR #67, opened 2022-03-17
Game initialization and lobby creation

VictorSchneuwly:
The game settings activity is now functional and gets the values needed to create a game.
The lobby activity is started and displays the information.


BartLucien (approved):
It seems good to me

app/src/main/java/com/github/fribourgsdp/radio/GameSettingsActivity.kt
```diff
@@ -1,11 +1,126 @@
...
+            selectedPlaylist = parent.getItemAtPosition(position).toString()
+            playlistSearchView.setQuery(selectedPlaylist, true)
+            startButton.isEnabled = true
+        }
+
+        startButton.setOnClickListener(startButtonBehavior())
+
+    }
+
+    private fun getHost() : User {
```
sim0mo, line 66: Ok, so we shall not forget to do that one of these days.

app/src/main/java/com/github/fribourgsdp/radio/GameSettingsActivity.kt
```diff
@@ -1,11 +1,126 @@
 package com.github.fribourgsdp.radio
 
+import android.content.Intent
 import androidx.appcompat.app.AppCompatActivity
 import android.os.Bundle
+import android.view.View
+import android.widget.*
+
+const val GAME_HOST_KEY = "com.github.fribourgsdp.radio.GAME_HOST"
```
sim0mo, line 9: Here it would be possible to store them in the api.xml, since this class inherits from Activity, and therefore has a context. This was not possible in LyricsGetter.

sim0mo (approved):
Good work, these activities look very good.

app/src/main/java/com/github/fribourgsdp/radio/GameSettingsActivity.kt
VictorSchneuwly: I looked up the kotlin documentation and they do it this way for intent keys.
What I was saying last time about your PR was about an api key given by the api creator.

PR #79, opened 2022-03-18
Correct keystore

Flamantrose321:
Remove the keystore from the public repo


VictorSchneuwly (approved):
It is indeed good to remove it from the repo.

nathanduchesne (approved):

PR #81, opened 2022-03-21
Serializable Data Classes

shanzhaiii:


shanzhaiii:
> Why are you commenting some tests ?

Because I have made Song names and artist names immutable as it makes more sense for serializable data structures

app/src/main/java/com/github/fribourgsdp/radio/User.kt
```diff
@@ -1,33 +1,31 @@
...
     private var linkedSpotify: Boolean = false
     val initial get(): Char = name.elementAt(0)
     val spotifyLinked get(): Boolean = linkedSpotify
 
-
-    private fun nameSanitizer(name: String) {
-        if (name.isEmpty() || name.isBlank()) {
-            throw  IllegalArgumentException("name cannot be empty or blank string")
+    companion object {
+        fun load(path: String) : User {
```
Flamantrose321, line 21: When you say path, do you mean the User IDs in Firebase or something else?

shanzhaiii: no, this is strictly for saving data locally on the device, so it is the relative path for the app specific storage

shanzhaiii: this will become useful in the save_user branch but it has been split into two PRs for clarity

Flamantrose321 (approved):
He answered my questions so everything is ok.

VictorSchneuwly:
> > Why are you commenting some tests ?
> 
> Because I have made Song names and artist names immutable as it makes more sense for serializable data structures. These tests were specifically testing parameter mutability.

But isn't it better to simply delete them? Since we probably won't make it mutable again.

shanzhaiii:
> 

Ok I will delete them

shanzhaiii:
> > > Why are you commenting some tests ?
> > 
> > 
> > Because I have made Song names and artist names immutable as it makes more sense for serializable data structures. These tests were specifically testing parameter mutability.
> 
> But isn't it better to simply delete them? Since we probably won't make it mutable again.

BartLucien (approved):
It is good for me

PR #83, opened 2022-03-23
add saving and loading capabilites to User

shanzhaiii:


app/src/main/java/com/github/fribourgsdp/radio/User.kt
```diff
@@ -28,6 +35,11 @@ class User (val name: String, val color: Int) {
                 Random.nextInt(100, 200)
     }
 
+    fun save(context: Context){
+        val userFile = File(context.filesDir, USER_DATA_PATH)
+        userFile.writeText(Json.encodeToString(this))
+    }
+
```
nathanduchesne, line 42: How persistent is this storage? Will this saved user state be kept when the app is closed?
All good for me!

nathanduchesne (approved):

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
```diff
@@ -25,10 +22,10 @@ class MainActivity : AppCompatActivity() {
...
         val profileButton: ImageButton = findViewById(R.id.profileButton)
-        profileButton.setOnClickListener {val intent : Intent = Intent(this, UserProfileActivity::class.java).apply {
-                putExtra(USERNAME, "Default")
-            }
-            startActivity(intent)
+        profileButton.setOnClickListener {
+            startActivity(Intent(this, UserProfileActivity::class.java))
         }
+
+        User("bla", User.generateColor()).save(this)
```
sim0mo, line 29: Is this line still necessary ?

app/src/main/java/com/github/fribourgsdp/radio/User.kt
shanzhaiii: yes, it is called app-specific storage: it is stored locally in the device but only accessible to your own app

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
shanzhaiii: this line is a way to demo that saving and loading works because the profile page now fetches from device storage. It can be replaced by anyone working on the demo as this PR is purely for adding backend features

sim0mo: Ok great !

sim0mo (approved):

PR #84, opened 2022-03-23
Extracting user playlists to transform them to our Playlist Data structure

nathanduchesne:
If applied, this commit will allow us to add playlists to a user to later on store them on our database.


nathanduchesne:
Please note this is not ready to be merged yet I still need to write unit tests. (Using manual testing I've seen the code already has the expected behaviour).

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
```diff
@@ -3,7 +3,8 @@ package com.github.fribourgsdp.radio
 import kotlinx.serialization.Serializable
 
 @Serializable
-data class Playlist (var name: String, var genre: Genre){
+data class Playlist (var name: String, var genre: Genre): java.io.Serializable{
```
shanzhaiii, line 6: do you really need to implement java.io.Serializable instead of directly using the Json parser like I have done in the user saving and loading? I think we should choose a single parsing library and use it

app/src/main/java/com/github/fribourgsdp/radio/Song.kt
```diff
@@ -5,7 +5,8 @@ import kotlinx.serialization.Serializable
 import java.util.concurrent.CompletableFuture
 
 @Serializable
-class Song (private val rawName: String, private val rawArtist: String, var lyrics: String) {
+class Song (private val rawName: String, private val rawArtist: String, var lyrics: String): java.io.Serializable {
+
```
shanzhaiii, line 9: same issue as in playlist

shanzhaiii (commented):
I am still wondering if we cannot uniformize the parsing lilbraries we are using

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
shanzhaiii: (this is still an import, just with kotlin syntax instead of java syntax)

shanzhaiii (approved):
After comparison with Kotlin's type of Serialize, the Serialize implemented in this branch produces cleaner code for intents so we will need to refactor to remove the other serializer. Approved

BartLucien (approved):
Looks good to me

PR #85, opened 2022-03-23
Strings

sim0mo:


app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
```diff
@@ -0,0 +1,98 @@
...
+class StringComparisons {
+    companion object{
+        private const val chars = "abcdefghijklmnopqrstuvwxyz1234567890"
+
+        /**
+        * Function that returns a number of points depending on the number of mistakes in the comparison of the two provided strings.
+        */
+        fun compareAndGetPoints(actual : String, expected : String, maxPts : Int = 100) : Int {
+            val oneErrorPenalty = 0.9
+            val twoErrorsPenalty = 0.8
```
Flamantrose321, line 16: Don't you want to put these values in constant?

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
```diff
@@ -0,0 +1,98 @@
...
+                1 -> (oneErrorPenalty*maxPts).toInt()
+                2 -> (twoErrorsPenalty*maxPts).toInt()
+                else -> 0
+            }
+        }
+
+        /**
+        * Function that compares two strings and returns the number of differences between them.
+        * Returns -1 if the number of differences is bigger than 2, to avoid taking too much time.
+        */
```
Flamantrose321, line 28: Don't you think we could also return a value from an enum for clarity?

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
```diff
@@ -0,0 +1,98 @@
...
+        }
+        /**
+        * Recursive function that determines if two formatted strings are equal at a given level of tolerance.
+        */
+        private fun equal(s1 : String, s2 : String, tolerance: Int) : Boolean {
+            var result : Boolean = false
+            if (tolerance >= 0){
+                if (s1 == s2){
+                    result = result || true
+                }
```
Flamantrose321, line 55: Why don't you go back to true directly here?

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
```diff
@@ -0,0 +1,98 @@
...
+        private fun CharSequence.unaccent() : String{
+            val temp = Normalizer.normalize(this, Normalizer.Form.NFD)
+            return "\\p{InCombiningDiacriticalMarks}+".toRegex().replace(temp, "")
+        }
+        /**
+        * Recursive function that determines if two formatted strings are equal at a given level of tolerance.
+        */
+        private fun equal(s1 : String, s2 : String, tolerance: Int) : Boolean {
+            var result : Boolean = false
+            if (tolerance >= 0){
```
Flamantrose321, line 52: Can the tolerance really be smaller than 0?

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
sim0mo: To reduce the number of return points of the function. Indeed, codeclimate was not agreeing with this way of coding in a previous commit.

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
sim0mo: No you're right. If you prefer I can replace it with == to make it more clear.

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
sim0mo: Yeah it would be better.

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
sim0mo: But to avoid execution of the rest of the function, I can put a condition in the second if, you're right.

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
sim0mo: Maybe not an enum, because this method is made to return integers (the maximum number can be adjusted with a constant). But it can be good to return a constant.

Flamantrose321 (approved):
He answered or made the necessary corrections to answer my questions so it's ok.

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
VictorSchneuwly: So why not just do `result = true` ?

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
```diff
@@ -0,0 +1,99 @@
+package com.github.fribourgsdp.radio
```
BartLucien, line 1: I think the part that compare the string and the part that interpret the comparison should be two different class

BartLucien (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
```diff
@@ -0,0 +1,99 @@
...
+                    result = result || true
+                }
+            }
+            if (tolerance >= 1 && !result){
+                if (s1.length == s2.length){
+                    result = result || (equalWithSubstitution(s1, s2, tolerance))
+                }else if (s1.length < s2.length){
+                    result = result || (equalWithDeletion(s1, s2, tolerance))
+                }else if (s1.length > s2.length){
+                    result = result || (equalWithInsertion(s1, s2, tolerance))
```
VictorSchneuwly, line 64: Also in this `if`, `result` will always be `false`. So doing `false || x` will always be equal to `x`. You could then simply assign the values to `result`.

VictorSchneuwly (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/StringComparisons.kt
sim0mo: I had left it to give an insight of what could be done at the very beginning, but it is not useful right now. So we can delete it.

BartLucien: Okay Okay

BartLucien (approved):
It is good for me

VictorSchneuwly (approved):
It's good for me now.

PR #87, opened 2022-03-24
Link firebase login to user

Flamantrose321:


app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
```diff
@@ -26,6 +26,5 @@ class MainActivity : AppCompatActivity() {
             startActivity(Intent(this, UserProfileActivity::class.java))
         }
 
-        User("bla", User.generateColor()).save(this)
```
shanzhaiii, line 29: this is still necessary for master as it will be used in the demo. You can remove it and I will add it back in my PR

shanzhaiii (approved):
A much needed update to the tests which will hopefully help cirrus fail less

sim0mo (approved):
Works fine for me.

PR #88, opened 2022-03-24
Extend database functionnalities

BartLucien:
Add some function in the database and refactor the database so that there is a localDatabase in order to run test without accessing FireStore.


VictorSchneuwly:
Maybe you should add some `@return` to the doc of every method that returns a `Task`, even `Task<Void>`, so that people using the Database will know they can use it.

app/src/main/java/com/github/fribourgsdp/radio/LocalDatabase.kt
```diff
@@ -0,0 +1,44 @@
...
+    override fun registerPlaylist(playlist: Playlist): Task<Void> {
+        val titleList : MutableList<Song> = mutableListOf()
+        for( song in playlist.getSongs()){
+            titleList.add(Song(song.name,song.artist,""))
+        }
+        playlistMap[playlist.name] = Playlist(playlist.name,titleList.toSet(),playlist.genre)
+
+        return Tasks.forResult(null)
+    }
+}
```
nathanduchesne, line 44: Why does this contain exactly the same code as the LocalDatabase class in our directory with the rest of our backend source code? This seems to cause problems with regard to code coverage, since this "mock implementation is fully tested, but the one in our main directory is not covered at all..

nathanduchesne (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/LocalDatabase.kt
BartLucien: I just deleted this one and the other one is in a separate package for better organisation of the project structure

nathanduchesne (approved):

Flamantrose321 (approved):
Update needed to avoid making calls to firebase for testing.

PR #106, opened 2022-03-30
Add playlists manually

sim0mo:
This PR will add the functionality to add playlists manually, out of any spotify connection.

### DIFF
--- a/app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
@@ -22,11 +22,7 @@ class AddPlaylistActivity : AppCompatActivity() {
         listView.adapter = listAdapter
         errorTextView = findViewById(R.id.addPlaylistErrorTextView)
 
-        user = try { /* TODO replace with proper error handling of asking user enter his info */
-            User.load(this)
-        } catch (e: java.io.FileNotFoundException) {
-            User("No User Found", User.generateColor())
-        }
+        user = UserProfileActivity.loadUser(this)
     }
     fun addItems(view : View) : Unit {
         val songNameTextView : EditText = findViewById(R.id.addSongToPlaylistSongName)
--- a/app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
@@ -12,10 +12,8 @@ class AddPlaylistActivity : AppCompatActivity() {
     private val listSongs = ArrayList<Song>()
     private var listNames = ArrayList<String>()
     lateinit var listAdapter: ArrayAdapter<String>
-    lateinit var spinnerAdapter: ArrayAdapter<Genre>
     lateinit var listView : ListView
     lateinit var errorTextView : TextView
-    lateinit var genreSpinner : Spinner
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
@@ -24,8 +22,6 @@ class AddPlaylistActivity : AppCompatActivity() {
         listView = findViewById(android.R.id.list)
         listView.adapter = listAdapter
         errorTextView = findViewById(R.id.errorTextView)
-        genreSpinner = findViewById(R.id.genreSpinner)
-        spinnerAdapter = ArrayAdapter(this, R.layout.playlist_item, Genre.values())
     }
     fun addItems(view : View) : Unit {
         val songName : String = findViewById<EditText>(R.id.songName).text.toString()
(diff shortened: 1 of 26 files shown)


app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -75,11 +79,18 @@ class UserProfileActivity : AppCompatActivity(), PlaylistAdapter.OnPlaylistClick
             text = if (user.spotifyLinked) "linked" else "unlinked"
         }
 
+        if (intent.getBooleanExtra(COMING_FROM_ADD_PLAYLIST_ACTIVITY_FLAG, false)){
+            addManualPlaylistToUser()
+            user.save(this)
+        }
+
```
shanzhaiii, line 86: I think this is a pretty ugly way of doing this. I didn't realise this was how @nathanduchesne had implemented it for the spotify import. I think it makes much more sense to save the playlists directly in AddPlaylistsActivity by loading the user (works in any activity), changing its playlists, and saving the user (works in any activity). This completely bypasses the heavy Intent and this ugly check of origin activity. It also makes more sense for UserProfileActivity to just stupidly load the userProfile everytime and not have to handle special events. I will need to do a PR for switching to a single serializable implementation so I will probably remove the other instance of saving playlists through intents in that PR

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -104,6 +115,11 @@ class UserProfileActivity : AppCompatActivity(), PlaylistAdapter.OnPlaylistClick
         user.addPlaylists(playlists)
     }
 
+    private fun addManualPlaylistToUser(){
+        val playlist = intent!!.getSerializableExtra("playlists") as HashSet<Playlist>
+        user.addPlaylists(playlist)
+    }
+
```
shanzhaiii, line 122: See my other comment

shanzhaiii (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
sim0mo: Okay, I'm going to refactor this. I will also modify the part from ImportSpotifyPlaylists, it might significantly simplify the intents.

app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
```diff
@@ -0,0 +1,59 @@
...
+        } else {
+            val intent = Intent(this@AddPlaylistActivity, UserProfileActivity::class.java)
+            val playlist = Playlist(playlistName, listSongs.toSet(), Genre.NONE)
+            // TODO: Add feature to select genre
+            user.addPlaylist(playlist)
+            user.save(this)
+            startActivity(intent)
+        }
+    }
+}
```
nathanduchesne, line 59: Great job! 
One feature we could add later on is a "fool proof mechanism": when the user presses on the back button and hasn't yet pressed on the create playlist button, we could intercept this and display something like "Are you sure you want to quit? If so everything you have done until now will not be saved" to make sure this situation does not occur.

nathanduchesne (approved):

app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
sim0mo: Oh yes, let's schedule this for another sprint !

app/src/main/res/values/strings.xml
```diff
@@ -40,4 +43,10 @@
     <string name="join_button_text">Join</string>
     <string name="join_fail_format">Failed to join lobby %d.</string>
     <string name="lobby_not_found">Lobby %d not found.</string>
+    <string name="add_a_playlist_manually">Add a playlist manually</string>
+    <string name="add_new_item">Add New Item</string>
```
Flamantrose321, line 47: Have you forgotten to use it for a content description?

app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
```diff
@@ -0,0 +1,59 @@
...
+        super.onCreate(savedInstanceState)
+        setContentView(R.layout.activity_add_playlist)
+        listAdapter = ArrayAdapter(this, android.R.layout.simple_list_item_1, listNames)
+        listView = findViewById(android.R.id.list)
+        listView.adapter = listAdapter
+        errorTextView = findViewById(R.id.addPlaylistErrorTextView)
+
+        user = UserProfileActivity.loadUser(this)
+    }
+    fun addItems(view : View) : Unit {
```
Flamantrose321, line 27: Do you plan to add tests to check that users are entering songs that actually exist?

sim0mo: No, since if they don't write songs that exist, they will be able to enter their own lyrics (if musixmatch does not find them). That's Alex' task for this week. Adding song verification is another story.

app/src/main/res/values/strings.xml
sim0mo: Bien-vu, I forgot to remove it.

Flamantrose321 (approved):
You have answered my questions and that seems correct.

PR #107, opened 2022-03-30
Load user instead of creating fake ones when joining or hosting lobbies.

VictorSchneuwly:
> Linked to task #107

Load the user from the phone instead of creating fake ones when joining or hosting lobbies.


VictorSchneuwly:
In order to make this work, we need to first merge the pull request of @BartLucien considering the update of users.

app/src/main/java/com/github/fribourgsdp/radio/User.kt
```diff
@@ -149,4 +149,12 @@ data class User (var name: String, val color: Int) {
     fun getSpotifyPlaylistUId(playlistName: String): String? {
         return playlistNamesToSpotifyId[playlistName]
     }
+
```
sim0mo, line 152: Good idea to use inheritance !

BartLucien (approved):
It looks good to me, as we talked IRL, it "just" replace the the name with the ID of the user everywhere it is needed and load the user instead of creating a fake one.

Flamantrose321 (approved):
Everything is good for me. It is indeed much more logical to identify a user by his id which is unique whereas several users can have the same name.

PR #109, opened 2022-03-31
Make user profile editable

BartLucien:
It makes the user profile editable and add a flow to register user.


app/src/androidTest/java/com/github/fribourgsdp/radio/MainActivityTest.kt
```diff
@@ -74,7 +74,7 @@ class MainActivityTest {
     fun correctTransitionToFireBaseTestActivity(){
```
Flamantrose321, line 74: The name of the test is no longer in line with what the test does.

app/src/main/java/com/github/fribourgsdp/radio/FirestoreDatabase.kt
```diff
@@ -125,6 +125,29 @@ class FirestoreDatabase : Database {
         }
     }
 
+    override fun generateUserId() : Task<Long> {
```
Flamantrose321, line 128: What is the use of this method knowing that when we connect to firebase we already have a unique id?

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
```diff
@@ -27,5 +33,28 @@ class MainActivity : AppCompatActivity() {
...
+            //if this is successful, we don't need to do anything
+            val user = User.load(this)
+
+        } catch (e: java.io.FileNotFoundException) {
+            createUser()
+        }
+
+
+    }
+    private fun createUser(){
```
Flamantrose321, line 48: Are you sure it makes sense to create a user when you haven't logged in yet? The sign-in to firebase is useless with this.

app/src/main/java/com/github/fribourgsdp/radio/CreateUserProfileActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+            startActivity(Intent(this, UserProfileActivity::class.java))
+
+        }
+
+
+
+    }
+    private fun prefillFields(){
+       // user = User.load(this)
+        //nameInput.setText(user.name)
```
nathanduchesne, line 45: Why does prefillFields do nothing?
If this is due to a problem getting the user, we could imagine writing dummy values (ex. John Doe) as prefilled fields for now.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -101,15 +162,19 @@ class UserProfileActivity : AppCompatActivity(), PlaylistAdapter.OnPlaylistClick
...
-        if(firebaseUser == null){
-            //user not logged in
-            startActivity(Intent(this, GoogleSignInActivity::class.java))
-            finish()
-        }else{
-            //user logged in
-            //get user info
-            /* TODO MAP FIREBASE INFO TO USER*/
-
+        if(firebaseUser != null){
```
Flamantrose321, line 165: It makes more sense to do this in google sign in where you can retrieve this information directly at login.

app/src/main/java/com/github/fribourgsdp/radio/CreateUserProfileActivity.kt
```diff
@@ -0,0 +1,48 @@
...
+        prefillFields()
+        saveButton.setOnClickListener(saveButtonBehaviour())
+    }
+
+    private fun saveButtonBehaviour() : View.OnClickListener{
+        return View.OnClickListener {
+            val textValue = nameInput.text
+            user = User(textValue.toString())
+            db.setUser("testCreateUserProfile",user)
+            user.save(this)
```
nathanduchesne, line 35: This may be a problem because it recreates a new user and will overwrite all information previously stored on the device in the case where a user has loaded his playlists for example.
A better solution would be to make a User's name property mutable, and use a try-catch here which would do:
try{
user.load
user.setName(textValue.toString())
}
catch {
what you have written now
}

app/src/main/java/com/github/fribourgsdp/radio/GoogleSignInActivity.kt
```diff
@@ -100,8 +100,8 @@ class GoogleSignInActivity : AppCompatActivity() {
             //check if user is new or existing
             if (authResult.additionalUserInfo!!.isNewUser) {
                 //user is new - Account Create
-                val mail = saveUserInDatabse(authResult)
-                saveTestUser(this, mail)
+               // val mail = saveUserInDatabse(authResult)
+                //saveTestUser(this, mail)
```
nathanduchesne, line 104: Why has this been removed? Won't this disable @Flamantrose321 's feature of having the newly logged in user's initials in his/her profile picture?

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
```diff
@@ -2,18 +2,24 @@ package com.github.fribourgsdp.radio
...
 import android.widget.ImageButton
 
 class MainActivity : AppCompatActivity() {
+    private val db = FirestoreDatabase()
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
         val fireBaseButton = findViewById<Button>(R.id.FireBaseButton)
-        fireBaseButton.setOnClickListener{startActivity(Intent(this,FireBaseTestActivity::class.java))}
+        fireBaseButton.setOnClickListener{startActivity(Intent(this,CreateUserProfileActivity::class.java))}
```
nathanduchesne, line 22: Please change this button name :)

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -36,31 +50,44 @@ class UserProfileActivity : AppCompatActivity(), PlaylistAdapter.OnPlaylistClick
...
 
-        user = loadUser(this)
+        instantiateViews()
+        try {
+             user = User.load(this)
+        } catch (e: java.io.FileNotFoundException) {
+            //this should never happen as a user is created and saved in the first activity
+            createDefaultUser()
+        }
+
```
nathanduchesne, line 61: I think the User should be loaded before instantiating views because if it isn't, certain displays will fail, such as the users playlists for example.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -92,6 +118,41 @@ class UserProfileActivity : AppCompatActivity(), PlaylistAdapter.OnPlaylistClick
...
+        user.id = id
+        val firebaseUser = firebaseAuth.currentUser
+        if(firebaseUser == null){
+            db.setUser(id,user)
+        }else{
+            db.setUser(firebaseUser.uid,user)
+        }
+
+        user.save(this)
+        usernameInitialText.setText(user.initial.uppercaseChar().toString())
```
nathanduchesne, line 154: I am pretty sure that this will not update the user.. Since l.144 you call instantiate a new User, it overwrites the one currently on local storage, which deletes all the prior information we had about that user. Hence it is more of an overwrite than an update.

nathanduchesne (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/FirestoreDatabase.kt
BartLucien: It is in case someone doesn't want to connect with google. He gets an id just for the session and can still play with his friends. The difference with a user that registered with google is that he won't have his data saved on the cloud.

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
BartLucien: Yes, this way, a user doesn't need to connect in order to play. When a user comes in, he gets a generated user that allows him to join a game. The sign-in to firebase is not useless, as it allows a user to connect to multiple phones and to store its data in the cloud

app/src/main/java/com/github/fribourgsdp/radio/CreateUserProfileActivity.kt
BartLucien: It is a relique of the process to get to the current form of the class. It will be removed.

app/src/main/java/com/github/fribourgsdp/radio/CreateUserProfileActivity.kt
BartLucien: This activity will be deleted anyway. This activity was to test the features I was going to implement on the UserProfileActivity.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
BartLucien: Yes it is. I still have, to transfer the playlist from the old_user to the new. As discussed IRL, I will just set the user.name as a variable in order to be able to just change it.

app/src/main/java/com/github/fribourgsdp/radio/GoogleSignInActivity.kt
BartLucien: See my answer to @Flamantrose321 as why I don't think we should do this here.

app/src/main/java/com/github/fribourgsdp/radio/MainActivity.kt
BartLucien: This was solved when merging from master.

app/src/androidTest/java/com/github/fribourgsdp/radio/MainActivityTest.kt
BartLucien: This was solved when merging from master.

app/src/androidTest/java/com/github/fribourgsdp/radio/UserProfileActivityTest.kt
```diff
@@ -2,29 +2,113 @@ package com.github.fribourgsdp.radio
...
+        ActivityScenario.launch<UserProfileActivity>(intent).use { scenario ->
+
+            onView(withId(R.id.username)).
+            check(matches(withText(testUser.name)));
+
+
+        }
+        Intents.release()
+
+    }*/
```
VictorSchneuwly, line 58: If you don't plan on using this test, you should delete this. If not, you should add a comment explaining why you commented all of this.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -92,6 +119,39 @@ class UserProfileActivity : AppCompatActivity(), PlaylistAdapter.OnPlaylistClick
...
+        saveChangeButton = findViewById(R.id.saveUserButton)
+        usernameInitialText = findViewById(R.id.usernameInitial)
+        spotifyStatusText = findViewById(R.id.spotifyStatus)
+        googleSignInButton = findViewById(R.id.googleSignInButton)
+        playlistDisplay = findViewById(R.id.playlist_recycler_view)
+        userIcon = findViewById(R.id.userIcon)
+    }
+
+    //this is theoritecally never called
+    private fun createDefaultUser(){
```
VictorSchneuwly, line 136: If it is never called, why keeping it?

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -36,31 +50,45 @@ class UserProfileActivity : AppCompatActivity(), PlaylistAdapter.OnPlaylistClick
...
         setContentView(R.layout.activity_user_profile)
 
-        firebaseAuth = FirebaseAuth.getInstance()
-        checkUser()
+        try {
+            user = User.load(this)
+        } catch (e: java.io.FileNotFoundException) {
+            //this should never happen as a user is created and saved in the first activity
+            createDefaultUser()
+        }
```
VictorSchneuwly, line 58: If you are sure this will never happen, I think you can remove the try catch. Or throw in the catch an UnexpectedException or something like that.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
VictorSchneuwly: I don't fully understand what would be the main difference. Could you give a small explanation @BartLucien?

VictorSchneuwly (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
BartLucien: My thought was that even though it should never happen, it could for reason we don't control. Better be safe than sorry.

app/src/androidTest/java/com/github/fribourgsdp/radio/UserProfileActivityTest.kt
BartLucien: Same as below

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
BartLucien: Same as below

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
VictorSchneuwly: @BartLucien explained this to me in real life and I agree.

VictorSchneuwly (approved):

app/src/androidTest/java/com/github/fribourgsdp/radio/PlaylistRecyclerViewTest.kt
```diff
@@ -21,9 +21,6 @@ import java.util.concurrent.TimeUnit
 
 @RunWith(AndroidJUnit4::class)
 class PlaylistRecyclerViewTest {
-    @get:Rule
-    var activityRule = ActivityScenarioRule(UserProfileActivity::class.java)
-    private lateinit var user: User
```
VictorSchneuwly, line 26: Why are you deleting this?

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -76,29 +76,22 @@ class UserProfileActivity : AppCompatActivity() {
...
-            startActivity(Intent(this,GoogleSignInActivity::class.java))
-        }
-
-        userIcon.setColorFilter(PorterDuffColorFilter(user.color, PorterDuff.Mode.ADD))
-
-        findViewById<ImageView>(R.id.userIcon).colorFilter =
-            PorterDuffColorFilter(user.color, PorterDuff.Mode.ADD)
+            if(signedIn){
+                firebaseAuth.signOut()
+                googleSignInButton.setText("Sign in")
```
VictorSchneuwly, line 86: Here, it would be better to use the resources instead of the hardcoded strings.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -112,12 +105,12 @@ class UserProfileActivity : AppCompatActivity() {
         firebaseAuth = FirebaseAuth.getInstance()
         usernameField = findViewById(R.id.username)
         launchSpotifyButton = findViewById(R.id.launchSpotifyButton)
-        logoutButton = findViewById(R.id.logoutButton)
+        logoutButton = findViewById(R.id.homeButton)
```
VictorSchneuwly, line 108: Here, you renamed the ID but not the variable.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -112,12 +105,12 @@ class UserProfileActivity : AppCompatActivity() {
...
         saveChangeButton = findViewById(R.id.saveUserButton)
         usernameInitialText = findViewById(R.id.usernameInitial)
         spotifyStatusText = findViewById(R.id.spotifyStatus)
         googleSignInButton = findViewById(R.id.googleSignInButton)
-        playlistDisplay = findViewById(R.id.playlist_recycler_view)
+        //playlistDisplay = findViewById(R.id.playlist_recycler_view)
```
VictorSchneuwly, line 113: Why are you commenting this line?

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -162,6 +155,9 @@ class UserProfileActivity : AppCompatActivity() {
                     user.save(this)
                 }
                 usernameField.setText(user.name)
+                googleSignInButton.setText("Sign out")
```
VictorSchneuwly, line 158: Also here, maybe not use hardcoded strings.

VictorSchneuwly (changes requested):

app/src/androidTest/java/com/github/fribourgsdp/radio/PlaylistRecyclerViewTest.kt
BartLucien: Because if we are using the rule and then launching the activity manually, it launches the activity twice, which slows up the execution of the test and maybe also make it less stable.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
BartLucien: Done

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
BartLucien: Done

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
BartLucien: Because this line became useless due to the new behaviour of the recycler view.

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
BartLucien: Done

VictorSchneuwly (approved):
Seems good to me now!

nathanduchesne (approved):

PR #110, opened 2022-03-31
Refactor java.io.serializable

shanzhaiii:
Important Pull Request to only use a single type of serialization in the project.
Please use Kotlin Serialization in the future with the @Serializable decorator instead of implementing Java.io.serializable
The choice has been made to keep Kotlin's implementation of Serializable instead of Java's because it makes writing to files much easier.

If you wish to use a Serializable class with Intents, you simply need to convert the instance to a string and pass it as an Extra string to the intent.
Example:
 val intent = Intent(this, PlaylistDisplayActivity::class.java)
            .putExtra(PLAYLIST_DATA, Json.encodeToString(userPlaylists[position]))
/////
Json.decodeFromString(intent.getStringExtra(PLAYLIST_DATA)!!)


BartLucien (approved):
This is fine for me.

VictorSchneuwly (approved):

PR #111, opened 2022-03-31
Provisionnaly fix JSON parsing bug on spotify import

sim0mo:
This PR, if applied, will fix a small bug when parsing playlist that contain non-JSON-parsable data. It will have to be done in a more elegant way one of these days.


app/src/main/java/com/github/fribourgsdp/radio/ImportSpotifyPlaylistsActivity.kt
```diff
@@ -22,6 +24,7 @@ class ImportSpotifyPlaylistsActivity : AppCompatActivity() {
         setContentView(R.layout.activity_import_spotify_playlists)
         val userToken = intent.getStringExtra("auth_token")
         TOKEN = userToken
+        Log.println(Log.INFO, "***", TOKEN.toString())
```
VictorSchneuwly, line 27: Do we need the `print` on the `Log` to stay for `master`?

app/src/main/java/com/github/fribourgsdp/radio/ImportSpotifyPlaylistsActivity.kt
```diff
@@ -113,17 +116,19 @@ class ImportSpotifyPlaylistsActivity : AppCompatActivity() {
...
                     val songs = parsedResponse.getJSONArray("items")
                     for (i in 0 until songs.length()){
-                        var songObject = songs.getJSONObject(i).getJSONObject("track")
+                        val songObject =
+                        try {
+                            songs.getJSONObject(i).getJSONObject("track")
+                        } catch (e : JSONException){
+                            // TODO: handle error
+                            continue
+                        }
```
VictorSchneuwly, line 129: Do you plan on doing this in a later sprint or did you forget about it?

app/src/main/java/com/github/fribourgsdp/radio/ImportSpotifyPlaylistsActivity.kt
sim0mo: No it is temporary.

app/src/main/java/com/github/fribourgsdp/radio/ImportSpotifyPlaylistsActivity.kt
sim0mo: No this was not handled at all previously, that is the purpose of this PR, to do a provisonal solution. Nathan may be responsible for it in another sprint.

BartLucien (approved):
It is fine for me

shanzhaiii (approved):
I approve this as a temporary solution but in general it is a good idea to start to implement error handling in what we are loading from spotify

PR #113, opened 2022-04-01
Playlist and song fragments

shanzhaiii:
This PR is important as it refactors the recyclerviews into fragments which could be easily used in other places. It also adds the ability to view a song's lyrics


shanzhaiii:
known bug: lyrics are saved at the right time but will not be updated in the song views until we return to MainActivity

VictorSchneuwly (approved):
Seems good to me

Flamantrose321 (approved):
nice to have fragments so that we can reuse recyclerviews in other activities

PR #126, opened 2022-04-02
Fix lyrics getter bug

sim0mo:
If applied, this pull request will fix a bug that made the app crash if the user would give an empty name as input for a song name. Indeed, function to cross out the name of the song in its lyrics was comparing strings in an error-prone way.
Furthermore, this PR fixes a large number a minor code style issues, for instance unused imports, _var_ instead of _val_ , hardcoded strings in activities, etc.


app/src/androidTest/java/com/github/fribourgsdp/radio/LyricsGetterTest.kt
```diff
@@ -102,22 +106,31 @@ class LyricsGetterTest {
...
...:1,\"has_subtitles\":1,\"has_richsync\":0,\"num_favourite\":0,\"album_id\":21057711,\"album_name\":\"The Bootleg Series, Vol. 5 - Bob Dylan Live 1975: The Rolling Thunder Revue\",\"artist_id\":13,\"artist_name\":\"Bob Dylan\",\"track_share_url\":\"https:\\/\\/www.musixmatch.com\\/lyrics\\/Bob-Dylan\\/Hurricane-Live?utm_source=application&utm_campaign=api&utm_medium=FribourgSDP%3A1409622437882\",\"track_edit_url\":\"https:\\/\\/www.musixmatch.com\\/lyrics\\/Bob-Dylan\\/Hurricane-Live\\/edit?utm_source=application&utm_campaign=api&utm_medium=FribourgSDP%3A1409622437882\",\"restricted\":0,\"updated_time\":\"2013-11-28T18:04:50Z\",\"primary_genres\":{\"music_genre_list\":[{\"music_genre\":{\"music_genre_id\":14,\"music_genre_parent_id\":34,\"music_genre_name\":\"Pop\",\"music_genre_name_extended\":\"Pop\",\"music_genre_vanity\":\"Pop\"}},{\"music_genre\":{\"music_genre_id\":21,\"music_genre_parent_id\":34,\"music_genre_name\":\"Rock\",\"music_genre_name_extended\":\"Rock\",\"music_genre_vanity\":\"Rock\"}},{\"music_genre\":{\"music_genre_id\":34,\"music_genre_parent_id\":0,\"music_genre_name\":\"Music\",\"music_genre_name_extended\":\"Music\",\"music_genre_vanity\":\"Music\"}}]}}}]}}}"
```
BartLucien, line 111: Could you just explain what is the purpose of these huge lines?

BartLucien (approved):
It is good for me

app/src/main/java/com/github/fribourgsdp/radio/GameSettingsActivity.kt
```diff
@@ -105,49 +105,46 @@ class GameSettingsActivity : AppCompatActivity() {
...
-                    playlistAdapter.filter.filter(query)
-                    selectedPlaylist = getPlaylist(query)
-                    startButton.isEnabled = true
-                    errorText.visibility = View.GONE
-                } else {
-                    errorText.text = getString(R.string.playlist_error_format, query)
-                    errorText.visibility = View.VISIBLE
-                    startButton.isEnabled = false
-                }
+    inner class SearchViewOnQueryBehavior: SearchView.OnQueryTextListener {
```
VictorSchneuwly, line 119: Does `inner class` make this class private? Because I don't think this should be accessed outside of `GameSettingsActivity`.

VictorSchneuwly: It is indeed accessible outside so you should make this private.

VictorSchneuwly (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/GameSettingsActivity.kt
VictorSchneuwly: Also, as you can read in the [doc for Inner classes](https://kotlinlang.org/docs/nested-classes.html#inner-classes), with the keyword `inner`, this class can access the members of its outer class. So it's better to write :
```kotlin
private class SearchViewOnQueryBehavior: SearchView.OnQueryTextListener { …
```

app/src/androidTest/java/com/github/fribourgsdp/radio/LyricsGetterTest.kt
sim0mo: This is just to _mock_ a response from musixmatch, to see if it is well interpreted. This was already there in my original pull request.
Here, it is indeed the response musixmatch gives you when you ask it to search for a song named "Hurricane" by "Bob Dylan".

app/src/main/java/com/github/fribourgsdp/radio/GameSettingsActivity.kt
sim0mo: Bien vu, I'll change that.

VictorSchneuwly (approved):
It's good for me now! It is indeed a better idea to instantiate the object this way rather than doing a method.

PR #128, opened 2022-04-05
Add playlists manually

sim0mo:
Small functionality for adding a genre selector when we create a new playlist.
Also updates list of genres.


Flamantrose321 (commented):
Nothing to report on the code.
Do you think in the future to add the possibility of having several genres for a playlist? It could be interesting

sim0mo:
> Nothing to report on the code. Do you think in the future to add the possibility of having several genres for a playlist? It could be interesting

Yes why not. Until now, we can use the NONE member of the enum, and it will display "no specific genre"

Flamantrose321 (approved):
All ok

BartLucien:
Do we also assume that the set of genre is fixed by us or that in the future, user can also set custom Genre to their playlist?

sim0mo:
> Do we also assume that the set of genre is fixed by us or that in the future, user can also set custom Genre to their playlist?

For now it is static, since an Enum is static. We should discuss if this it is something we want to implement, but for now it is not primordial.

BartLucien (approved):
All good to me

PR #129, opened 2022-04-05
Flexible playlist creation

sim0mo:
This PR will add the functionality to remove songs from playlists when we are creating them, to edit playlists and to delete playlists.


app/src/main/java/com/github/fribourgsdp/radio/PlaylistSongsFragment.kt
```diff
@@ -33,6 +39,24 @@ class PlaylistSongsFragment : MyFragment(R.layout.fragment_playlist_display), On
...
         initializeRecyclerView()
+        //sets listeners to 2 buttons
+        editButton = requireView().findViewById(R.id.editButton)
+        editButton.setOnClickListener {
+            val intent  = Intent(context, AddPlaylistActivity::class.java)
+            intent.putExtra(PLAYLIST_TO_MODIFY, Json.encodeToString(playlist))
+            startActivity(intent)
+        }
+        deleteButton = requireView().findViewById(R.id.deleteButton)
+        deleteButton.setOnClickListener{
```
BartLucien, line 50: I think that clicking on a playlist to delete is rather unintuitive, even if indicated. It think it would be more intuitive to be able to slide it to the side in order to delete it.

nathanduchesne: It may not be intuitive for you, but I think for many people of the "older" generation it is actually much more intuitive to first select the playlist and then delete it with the appropriate button. The risk with sliding is that people unintentionally delete their playlists.

One feature that could be added is the fact that when a user clicks on delete, a popup comes up and asks if he really wants to delete it. This double check could be a nice addition but is most definitely not needed for now.

nathanduchesne (approved):

app/src/main/java/com/github/fribourgsdp/radio/User.kt
```diff
@@ -112,6 +111,15 @@ data class User (var name: String, val color: Int) {
...
     }
 
+    /**
+     * removes a single playlist given by its name from the user's set of playlists
+     *
+     * @param name the name of the playlist to remove
+     */
+    fun removePlaylistByName(name: String) : Boolean{
+        return playlists.removeIf{p -> p.name == name}
+    }
```
Flamantrose321, line 121: If several playlists have the same name, they will all be deleted? If so, this is not consistent with the documentation of the function.

sim0mo: Oh yes you're right, I shall update the documentation. In fact, name is the only thing that defines a playlist locally, its "primary key" to speak in term of db. So yes it might delete multiple ones if they have same name

app/src/main/java/com/github/fribourgsdp/radio/PlaylistSongsFragment.kt
sim0mo: I see what you mean @BartLucien, i.e. sliding instead of clicking on _Songs_ to remove them, _once we are in the AddPlaylistActivity_ . This swiping could indeed be implemented.
The popup on playlist deletion is also a great idea, I will implement it in a future sprint.

Flamantrose321 (approved):

app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
```diff
@@ -3,31 +3,57 @@ package com.github.fribourgsdp.radio
...
+                listAdapter.notifyItemRemoved(position)
+            }
+        })
+
+        recyclerView.adapter = listAdapter
+        val layoutManager = LinearLayoutManager(this)
+        layoutManager.stackFromEnd = true
+        recyclerView.layoutManager = layoutManager
+        errorTextView = findViewById(R.id.addPlaylistErrorTextView)
+
```
shanzhaiii, line 50: Ideally this would use the recycler view fragment but I agree not using it here because of the specifics of format it takes. Possibly a later refactor

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
```diff
@@ -78,4 +78,12 @@ data class Playlist (var name: String, var genre: Genre) {
...
         return songs.toSet()
     }
+
+    override fun hashCode(): Int {
+        return name.hashCode()
+    }
+
+    override fun equals(other: Any?): Boolean {
+        return name == (other as Playlist).name
+    }
```
shanzhaiii, line 88: Useless because Playlist is a data class

app/src/main/java/com/github/fribourgsdp/radio/User.kt
shanzhaiii: Should not happen because I think playlists are in a set at least before being passed to recyclerviews

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -156,12 +152,20 @@ class UserProfileActivity : AppCompatActivity() {
...
         }
     }
+
+    override fun onBackPressed() {
+        super.onBackPressed()
+        val intent = Intent(this, MainActivity::class.java)
+        intent.putExtra(RECREATE_USER, false)
+        startActivity(intent)
+        finish()
+    }
```
shanzhaiii, line 168: Really not a fan of this as it can really mess with fragments when we don't intend for back to return to main menu. Is it still necessary? if we remove it, can we not just never create a user and create it in the try / catch of no use being found on load?

shanzhaiii (commented):
I have a couple non-critical remarks

app/src/main/java/com/github/fribourgsdp/radio/AddPlaylistActivity.kt
sim0mo: Exactly, I had the same idea

app/src/main/java/com/github/fribourgsdp/radio/Playlist.kt
sim0mo: Not so useless because we want to be able to exclude playlist from sets (HashSet/TreeSet) based only on their name, so that no two playlists with the same name are in that set.

shanzhaiii: Yes indeed you are absolutely right, my bad

PR #130, opened 2022-04-05
Clean project

BartLucien:
This PR should uniformize the code in the test and hopefully make them run faster. It also "corrects" some code by using the approriate function and move some function in classes where they should be defined instead of in activity directly.


app/src/androidTest/java/com/github/fribourgsdp/radio/AddPlaylistActivityTest.kt
```diff
@@ -104,7 +120,6 @@ class AddPlaylistActivityTest {
     }
```
sim0mo, line 120: I think you forgot to remove line 119 to release Intents as you do a _@After_ ;)
That might be the reason why tests are failing...

app/src/main/java/com/github/fribourgsdp/radio/ImportSpotifyPlaylistsActivity.kt
```diff
@@ -38,10 +39,12 @@ class ImportSpotifyPlaylistsActivity : AppCompatActivity() {
     companion object {
 
         fun saveInfoToUser(playlistNameToUId: HashMap<String, String>, playlists: Set<Playlist>, ctx : Context) {
-            val user = UserProfileActivity.loadUser(ctx)
-            user.addSpotifyPlaylistUids(playlistNameToUId)
-            user.addPlaylists(playlists)
-            user.save(ctx)
+            User.loadOrDefault(ctx).addOnSuccessListener { u ->
+                u.addSpotifyPlaylistUids(playlistNameToUId)
```
sim0mo, line 43: What happens if it fails ? Shouldn't it throw an Exception or something ?

app/src/androidTest/java/com/github/fribourgsdp/radio/HostGameHandlerTest.kt
```diff
@@ -37,7 +38,7 @@ class HostGameHandlerTest {
         handler.handleSnapshot(mockSnapshot)
 
         // Wait for the task of the database to execute
-        Thread.sleep(1000)
+        Thread.sleep(sleepingTime)
```
Flamantrose321, line 41: I didn't see it in the previous pull requests, but isn't there a way to do this more cleanly without having to put a Thread.sleep?

app/src/androidTest/java/com/github/fribourgsdp/radio/HostGameHandlerTest.kt
```diff
@@ -57,7 +58,7 @@ class HostGameHandlerTest {
...
```
Flamantrose321, line 61: Same as above

Flamantrose321 (commented):
Important update for the maintainability of the project and the cleanliness of the code :)

app/src/androidTest/java/com/github/fribourgsdp/radio/HostGameHandlerTest.kt
BartLucien: For this you have to see with @VictorSchneuwly but it seemed to be obligatory as a way to let the application have the time to do the things it needs to do. Locally, it works with 1 millisecond as sleepingTime, but it failed once on Cirrus so I increased the value a litlle bit.

Flamantrose321 (commented):
I may have the reason why you don't pass the tests.
By putting the mockActivity in the Android test folder, the app is no longer able to find them. As you should see in your Android Manifest. 
Indeed, there are 2 APK generated during project building. First is APK with application and second APK contains test. If you put your activity in test folder it will be in second APK that is used for test and your application APK does not contains it. Maybe that's why your tests don't pass.

Edit by Lucien: this was not the problem. As weird as it might seen, having the manifest buggy doesn't disturb the test.

Flamantrose321 (approved):
Everything is ok for me

sim0mo (approved):

PR #134, opened 2022-04-06
Voice over ip integration

Flamantrose321:
If applied, this PR will allow us to have voice channels in our game lobby.

Code climate signals a few code smells that will be fixed when we delete the voice over ip demo activity next week.


shanzhaiii (commented):
a couple small comments

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -41,15 +51,18 @@ class GameActivity : AppCompatActivity(), GameView {
...
         playerGameHandler.linkToDatabase()
 
     }
 
+    override fun onDestroy() {
+        super.onDestroy()
+        voiceChannel.leaveChannel()
+        RtcEngine.destroy()
+    }
+
```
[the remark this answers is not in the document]
nathanduchesne: Yes indeed, as long as the activity isn't destroyed, the players are in the voice chat. This is great because it allows pop-up fragments to appear (when the singer must choose a song) without interrupting the audio.

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
```diff
@@ -138,12 +152,23 @@ open class LobbyActivity : AppCompatActivity() {
         playersListView = findViewById(R.id.playersListView)
         namesAdapter = ArrayAdapter(this, android.R.layout.simple_list_item_1)
         playersListView.adapter = namesAdapter
+    }
 
+    private fun initButtons() {
         launchGameButton = findViewById(R.id.launchGameButton)
-
+        askForPermissionsButton = findViewById(R.id.micPermissionsButton)
```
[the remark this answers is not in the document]
nathanduchesne: What actually happens is that when we create the activity, we check for permissions, and if the user has already enabled microphone usage, this button does not display. 
If the permissions aren't active yet, then the user has to accept them. You are right we could have done it in another manner (could be a task for next week!)

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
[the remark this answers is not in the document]
shanzhaiii: makes sense thank you

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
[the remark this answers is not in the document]
shanzhaiii: Ok and so the voice chat cannot start until permission is given or can the application still crash?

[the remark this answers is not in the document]
Flamantrose321: In this version of the application, if the host does not have the permissions, it cannot start the game. And if it's not the host, when the host launches, players who don't have permissions are not added to the game.
We may change this in the future, but it is not a priority.

app/src/androidTest/java/com/github/fribourgsdp/radio/GameActivityTest.kt
```diff
@@ -93,6 +98,8 @@ class GameActivityTest {
         val testIntent = Intent(ctx, GameActivity::class.java).apply {
             putExtra(GAME_IS_HOST_KEY, true)
             putExtra(GAME_KEY, json.encodeToString(fakeGame))
+            putExtra(GAME_UID, 37L)
+            putExtra(IS_IN_TEST_MODE, true)
```
BartLucien, line 102: As discussed today when you said you were now using mockito, is this still needed?

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -1,13 +1,20 @@
...
 import android.view.inputmethod.EditorInfo
 import android.widget.*
-import androidx.appcompat.app.AppCompatActivity
+import io.agora.rtc.Constants
+import io.agora.rtc.RtcEngine
 import kotlinx.serialization.decodeFromString
 import kotlinx.serialization.json.Json
+import kotlin.random.Random
+
+const val IS_IN_TEST_MODE = "Test mode for voice chat engine"
```
BartLucien, line 17: Linking to previous comment, is it still needed?

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -131,6 +150,22 @@ class GameActivity : AppCompatActivity(), GameView {
...
+            voiceChannel.mute(muteButton)
+        }
+    }
+
+    private fun initVoiceChat(is_test_mode: Boolean, gameUid: Long) {
+
+
+        voiceChannel = VoiceIpEngineDecorator(this)
+
+        val userId = Random.nextInt(100000000)
```
BartLucien, line 164: As discussed, this will be changed with the id of the user once we will have a generated a full id without character also for user signed in with google?

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -32,6 +44,8 @@ class GameActivity : AppCompatActivity(), GameView {
 
         lateinit var playerGameHandler: PlayerGameHandler
 
+        initVoiceChat(intent.getBooleanExtra(IS_IN_TEST_MODE, true), intent.getLongExtra(
```
BartLucien, line 47: Same as other comment?

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
```diff
@@ -210,17 +235,33 @@ open class LobbyActivity : AppCompatActivity() {
         gameBuilder.setUserList(playersList.map { name -> User(name) })
     }
 
-    private fun goToGameActivity(isHost: Boolean, game: Game? = null, gameID: Long? = null) {
+    private fun goToGameActivity(isHost: Boolean, game: Game? = null, gameID: Long) {
         val intent: Intent = Intent(this, GameActivity::class.java).apply {
             putExtra(GAME_IS_HOST_KEY, isHost)
+            putExtra(IS_IN_TEST_MODE, false)
```
BartLucien, line 241: Same as above comment?

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
nathanduchesne: Yes indeed, for now the Google unique id for users has letters in it, but our voiceChannel uid requires an integer in Agora. This wouldn't be a problem if we only needed to feed it to Agora, but we need the function to be two-way to then extract which user is the loudest in the room chat. -> This should be done next week.

app/src/androidTest/java/com/github/fribourgsdp/radio/GameActivityTest.kt
nathanduchesne: You are right, we will remove it.

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
nathanduchesne: Yes indeed.

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
Flamantrose321: No it's an oversight thanks for pointing it out.

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
Flamantrose321: No it's an oversight thank you for pointing it out.

BartLucien (approved):
It is good for me

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -40,17 +52,22 @@ class GameActivity : AppCompatActivity(), GameView {
             val hostGameHandler = HostGameHandler(game, this)
             hostGameHandler.linkToDatabase()
             user = User("The best player")
-            playerGameHandler = PlayerGameHandler(game.id, this)
-        } else {
-            playerGameHandler = PlayerGameHandler(intent.getLongExtra(GAME_UID_KEY, -1L), this)
         }
+        playerGameHandler = PlayerGameHandler(intent.getLongExtra(GAME_UID_KEY, -1L), this)
```
VictorSchneuwly, line 56: Here and at line 46 you both times get the extra id of the intent. It should be better to only get it once and store it in a `val`. Also, the default shouldn't be 0 as it is a possible lobby/game value, so it's better to set it to minus one.

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -19,11 +25,16 @@ class GameActivity : AppCompatActivity(), GameView {
     private lateinit var songTextView : TextView
     private lateinit var errorOrFailureTextView : TextView
     private lateinit var songGuessEditText : EditText
+    private lateinit var muteButton : ImageButton
     private lateinit var songGuessSubmitButton: Button
 
     private lateinit var playersListView : ListView
     private lateinit var namesAdapter : ArrayAdapter<String>
 
+    private lateinit var playerGameHandler: PlayerGameHandler
```
VictorSchneuwly, line 34: The `playerGameHandler` is only used in `onCreate` so no need to extract it.

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
```diff
@@ -192,7 +217,7 @@ open class LobbyActivity : AppCompatActivity() {
 
                 val isGameLaunched = snapshot.getBoolean("launched")
 
-                if (!isHost && isGameLaunched != null && isGameLaunched) {
+                if (!isHost && isGameLaunched != null && isGameLaunched && hasVoiceIdPermissions) {
```
VictorSchneuwly, line 220: Here instead of just not doing anything, it would be better to display an error when `hasVoiceIdPermissions` is false. The other conditions of this if-statement do not trigger an error as it is made to redirect players.

app/src/androidTest/java/com/github/fribourgsdp/radio/GameActivityTest.kt
```diff
@@ -20,6 +20,8 @@ import org.junit.Test
 import org.junit.runner.RunWith
 
 
+
+
```
VictorSchneuwly, line 24: Next time don't use `git commit -am` ;)

VictorSchneuwly (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
nathanduchesne: You are right. What we wanted to implement as a feature for next week is that the host cannot launch a lobby until all the players have the appropriate permissions. For now we just wanted to make sure that a user wouldn't try initializing a voice channel and joining it without having the appropriate permissions.

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
nathanduchesne: No it is also called in initViews() :)

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
nathanduchesne: Good catch -> done!

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
nathanduchesne: Indeed, this has been corrected.

app/src/main/java/com/github/fribourgsdp/radio/LobbyActivity.kt
VictorSchneuwly: Ok it is even better this way. All good if you do it next week. 👍🏻

VictorSchneuwly (approved):
All good for me now, but we must not forget to correct the behavior of someone not accepting the conditions for next week 👍🏻

sim0mo (commented):
Tested it live with Maxence and it works well!

PR #140, opened 2022-04-08
Fix lyrics not updating

shanzhaiii:
Fixing the bug of lyrics not updating in fragments by reloading the data from local app storage on every fragment creation instead of using bundle input


shanzhaiii:
having to use addPlaylist instead of addPlaylists for test to work is unusual. needs investigation

shanzhaiii:
Test behaviour undefined. Converting FileSystem to a mockable interface

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockFileSystem.kt
```diff
@@ -0,0 +1,27 @@
+package com.github.fribourgsdp.radio.mockimplementations
```
BartLucien, line 1: Good idea, I will probably make use of this

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockFileSystem.kt
```diff
@@ -0,0 +1,27 @@
...
+
+import android.content.Context
+import com.github.fribourgsdp.radio.FileSystem
+
+var data: String = ""
```
BartLucien, line 6: Why don't you put this inside the class directly?

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockUserProfileActivity.kt
```diff
@@ -0,0 +1,28 @@
+package com.github.fribourgsdp.radio.mockimplementations
+
+import android.content.Context
+import android.os.Bundle
+import com.github.fribourgsdp.radio.*
+import org.mockito.Mockito
+import org.mockito.Mockito.mock
+
+const val userName = "test"
```
BartLucien, line 9: Why don't you put this in the class directly?

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockFileSystem.kt
shanzhaiii: Since it is a "filesystem", and my implementation generates a new file system instance every time we need to read or write due to backwards compatibility with the context signature, I need it to be a global variable so that the string is the same between filesystem instances

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockUserProfileActivity.kt
shanzhaiii: In the tests since the activity is initialized by an instance, i don't have access to that activites internals and I could not be bothered to make a companion object for a test but i can change it if you deem it is necessary

app/src/main/java/com/github/fribourgsdp/radio/User.kt
```diff
@@ -46,10 +42,14 @@ data class User (var name: String, val color: Int) {
 
 
     companion object {
+        const val USER_DATA_PATH = "user_data_file"
+        private var fileSystemGetter: FileSystem.FileSystemGetter =
```
nathanduchesne, line 46: If I understand correctly, this is the working FS implementation we previously used wrapped in a a class, such that you can mock it when needed in MockFileSystem right?
Even though it's not the most "logical" place to put this attribute considering a User shouldn't store the info that indicates where it is going to be stored, I understand it's by far the easiest way to do so, so it's fine for me.

app/src/main/java/com/github/fribourgsdp/radio/SetUtility.kt
```diff
@@ -15,4 +15,13 @@ object SetUtility {
     fun <E> addAllToSet(set : MutableSet<E>, items : Iterable<E>){
         items.forEach{ i : E -> addToSet(set, i)}
     }
+
+    fun <E : Nameable> getNamedFromSet(set : MutableSet<E>, name: String) : E {
+        val filteredSet = set.filter { nameable -> nameable.name == name }
+        if (filteredSet.isEmpty()) {
+            throw NoSuchElementException()
+        } else {
+            return filteredSet[0]
```
nathanduchesne, line 24: Are Nameable elements in a set uniquely defined by their name attribute? If so I see no problems here but if not, won't we lose some data by only taking the first elem?

app/src/main/java/com/github/fribourgsdp/radio/SongFragment.kt
```diff
@@ -60,6 +52,15 @@ open class SongFragment : MyFragment(R.layout.fragment_song) {
         val artistTitle : TextView = requireView().findViewById(R.id.ArtistName)
         val lyricsEditText : EditText = requireView().findViewById(R.id.editTextLyrics)
 
+        val user = User.load(requireContext())
+        playlist = user.getPlaylistWithName(playlistName)
+        song = playlist.getSong(songName)
+        initialLyrics = song.lyrics
+        currentLyrics = initialLyrics
+        if(song.lyrics == ""){
```
nathanduchesne, line 60: Shouldn't you replace this with your initialLyrics variable?

nathanduchesne (approved):

app/src/main/java/com/github/fribourgsdp/radio/User.kt
shanzhaiii: Yes. I am using a getter instead of the actual filesystem directly to preserve backwards compatibility with the context signature. I am forced to put it in the companion object instead of a class instance because it also needs to be used by the static load method. The USER_DATA_PATH variable has just been moved from a global variable to the companion object because I find it cleaner. We use this path because it is the default path for user files, but you could technically store several users if you choose several paths.

app/src/main/java/com/github/fribourgsdp/radio/SongFragment.kt
shanzhaiii: this is not my code but yes

app/src/main/java/com/github/fribourgsdp/radio/SetUtility.kt
shanzhaiii: hmmm you are right, this implementation will not work with identically named songs of different artists. I will change it next week to a better implementation

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockFileSystem.kt
BartLucien: Okay

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockUserProfileActivity.kt
BartLucien: No it is fine for me

BartLucien (approved):

PR #141, opened 2022-04-08
Make edited playlist savable.

sim0mo:
Bugfix : Replaces the behaviour of the addPlaylist function, so that when a playlist is added to the set of playlists of the user, it replaces the one that has the same name.

### DIFF
--- a/app/src/main/java/com/github/fribourgsdp/radio/SetUtility.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/SetUtility.kt
@@ -1,0 +1,18 @@
+package com.github.fribourgsdp.radio
+
+object SetUtility {
+    /**
+     * Add an item to a set, after removing it if it was present
+     */
+    fun <E> addToSet(set : MutableSet<E>, item : E){
+        set.removeIf { p : E -> p == item }
+        set.add(item)
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


shanzhaiii (changes requested):
If we make this change, it should be also done fore addPlaylists and the equivalent song methods

app/src/main/java/com/github/fribourgsdp/radio/SetUtility.kt
```diff
@@ -0,0 +1,18 @@
+package com.github.fribourgsdp.radio
+
+object SetUtility {
+    /**
+     * Add an item to a set, after removing it if it was present
+     */
+    fun <E> addToSet(set : MutableSet<E>, item : E){
```
Flamantrose321, line 7: Isn't there a function or class that does the same thing in the Kotlin documentation?

sim0mo: Unfortunately not, we have to remove and then add, and as we do this frequently, it was to avoid code duplication that I created this static class/object.

Flamantrose321 (approved):
ok for me

shanzhaiii (approved):
Much better now that it affects all similar data classes. Approved!

PR #154, opened 2022-04-13
Add Active User display

Flamantrose321:
This PR will allow players to see the name of the person who is speaking.

I've also added some tests for methods that were not yet tested.

### DIFF
--- a/app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
@@ -155,15 +155,12 @@ open class GameActivity : AppCompatActivity(), GameView, User.Loader {
 
     open protected fun initVoiceChat(gameUid: Long) {
 
-        val map = mapIdToName.mapKeys { it.hashCode().absoluteValue } + Pair(3, "Jean-Marie Le Pen")+Pair(4, "Tom Dumoulin")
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -153,12 +154,13 @@ open class GameActivity : AppCompatActivity(), GameView, User.Loader {
     }
 
     open protected fun initVoiceChat(gameUid: Long) {
-        voiceChannel = VoiceIpEngineDecorator(this, MyIRtcEngineEventHandler(this))
-        val userId = Random.nextInt(100000000)
+
+        val map = mapIdToName.mapKeys { it.hashCode().absoluteValue }
+        if (!this::voiceChannel.isInitialized) voiceChannel = VoiceIpEngineDecorator(this, MyIRtcEngineEventHandler(this, map))
+        val userId = user.name.hashCode().absoluteValue
```
sim0mo, line 160: Is that enough to guarantee uniqueness ?

sim0mo (approved):

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
Flamantrose321: Good point. Yes, that's enough.

VictorSchneuwly (approved):
Everything seems good to me!

PR #156, opened 2022-04-14
Disable the host's ability to launch the game when players are missing the necessary mic permissions

nathanduchesne:
If accepted this PR will implement the following features:

1. Disable the host's ability to launch before all players have accepted the microphone permissions on the rad.io app.
2. Have a user-friendly interface to display all user microphone permissions, such that all players know who is hasn't accepted the permissions yet.


app/src/androidTest/java/com/github/fribourgsdp/radio/FirestoreDatabaseTest.kt
```diff
@@ -27,4 +27,14 @@ class FirestoreDatabaseTest {
         assertEquals( null,Tasks.await(user))
     }
 
+    @Test
```
BartLucien, line 30: So far, we haven't tested this class, as since week 3, we avoid testing with API. If really we can keep this test, as we have sufficient API calls and google is rather a stable company, but in this case, set the id to a id that is not possibly generated by our application. If a user has Id 215 and this test run, troubles will occur.

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockLobbyActivities.kt
```diff
@@ -27,6 +27,10 @@ class WorkingLobbyActivity : LobbyActivity() {
 
         startActivity(intent)
     }
+
+    override fun loadUser(): User {
```
BartLucien, line 31: I think it would be a good idea to also set the id explicitly, as it may be used in the activity

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockLobbyActivities.kt
```diff
@@ -47,4 +51,8 @@ class BuggyLobbyActivity : LobbyActivity() {
...
+        return User("the buggy player")
```
BartLucien, line 56: Same as above

BartLucien (changes requested):
The part with the test modifying a possible valid value of a user has to change.

VictorSchneuwly (changes requested):
I think you need to put the file [`mic_no_permissions-playstore.png`](https://github.com/FribourgSDP/rad.io/blob/137-disable-game-launch-when-missing-permissions/app/src/main/mic_no_permissions-playstore.png) inside `res`.

Also, in the `RecyclerLobbyAdapter`, it could be a good idea to give a map user -> image instead of two separate lists. It would make the code easier to comprehend.

nathanduchesne:
> I think you need to put the file [`mic_no_permissions-playstore.png`](https://github.com/FribourgSDP/rad.io/blob/137-disable-game-launch-when-missing-permissions/app/src/main/mic_no_permissions-playstore.png) inside `res`.
> 
> Also, in the `RecyclerLobbyAdapter`, it could be a good idea to give a map user -> image instead of two separate lists. It would make the code easier to comprehend.

This is a good idea, but a map implies the keys must be unique, whereas we said the displayed usernames aren't necessarily unique.. If you want I can group the two methods together to have one "setContent" method that takes the images as well as the names, but a map won't do the job.

app/src/androidTest/java/com/github/fribourgsdp/radio/FirestoreDatabaseTest.kt
nathanduchesne: I've changed the test by creating a mock entry in our firestore DB, such that there is no risk the problems you stated may ever occur (the mock lobby entry has uid 123456789, whereas our lobby id's in game are always %1000).

BartLucien: It is fine for me for me. It will likely disappear in the future as I have managed to find the way the mock the database so you're test will probably be replaced with a proper test that doesn't do any API call.

BartLucien (approved):

VictorSchneuwly (approved):

PR #161, opened 2022-04-17
152 add playlist and song on the database

BartLucien:
If accepted, this PR will also store the song and the playlist of a user in the cloud. 
Note that it doesn't retrieve yet the song of a playlist, as it requires more work a bit everywhere in the codebase and also a reflection about what should be stored on the cloud and what not, and this implies the question of when to generate Id's for songs or not.


app/src/main/java/com/github/fribourgsdp/radio/Database.kt
```diff
@@ -51,7 +51,9 @@ interface Database {
      * @return a task loading a unique ID for the user.
      */
     fun generateUserId() : Task<Long>
+    fun generatePlaylistId() : Task<Long>
```
sim0mo, line 54: A short doc comment would be appreciated ;)

app/src/main/java/com/github/fribourgsdp/radio/Genre.kt
```diff
@@ -23,7 +23,7 @@ enum class Genre (private val genreName: String){
     MOVIE("Movie Music"),
     COUNTRY("Country");
 
-    override fun toString(): String {
+    /*override fun toString(): String {
```
sim0mo, line 26: I understand this is necessary the cloud, but could you then rename this function here and wherever it is called ? so we do not overwrite tostring but we do not remove a functionnality

app/src/main/java/com/github/fribourgsdp/radio/FirestoreDatabase.kt
```diff
@@ -1,31 +1,65 @@
...
+    constructor():this(FirestoreRef())
 
     override fun getUser(userId : String): Task<User> {
-        return  db.collection("user").document(userId).get().continueWith { l ->
+        return  refMake.getUserRef(userId).get().continueWith { l ->
             val result = l.result
             if(result.exists()){
-                User(result["first"].toString())
+                val userId = result["userID"].toString()
+                val playlists = result["playlists"] !! as ArrayList<HashMap<String,String>>
```
sim0mo, line 47: Have you considered the case where the result has null values for playlists and other fields, i.e. everywhere you use `!!` ?

BartLucien: I have not and I know it is kind of a strong assumption. But the only function setting a user is setUser, where we set these exact fields. It is not retro compatible but we don't need retro compatility toward an early phase of our project imo.

app/src/main/java/com/github/fribourgsdp/radio/Genre.kt
BartLucien: Yes of course

app/src/main/java/com/github/fribourgsdp/radio/Database.kt
BartLucien: Of course

app/src/main/java/com/github/fribourgsdp/radio/FirestoreDatabase.kt
sim0mo: Ok fair enough !

nathanduchesne (commented):
I know most of this code was already there previously, but I feel there are many inconsistencies between storing the user locally and on the cloud. I think one task for next week could be thinking about what data we want to store where, such that we avoid having useless redundancies. This will allow our UserProfileActivity to become a bit more clear and avoid having to go through the different combinations (ex. not on db/but local or the inverse) in checkUser().

app/src/main/java/com/github/fribourgsdp/radio/UserProfileActivity.kt
```diff
@@ -48,6 +50,7 @@ class UserProfileActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_user_profile)
         instantiateViews()
+        Log.d(ContentValues.TAG, "DocumentSnapshot added with ID put: do I come here 5" )
```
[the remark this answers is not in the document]
BartLucien: Yes of course

app/src/main/java/com/github/fribourgsdp/radio/FirestoreDatabase.kt
```diff
@@ -1,31 +1,65 @@
...
+    open fun getUserRef(userId : String) : DocumentReference{
+        return db.collection("user").document(userId)
+    }
+    open fun getSongRef(songId : String) : DocumentReference{
+        return db.collection("songs").document(songId)
+    }
+    open fun getPlaylistRef(playlistId : String) : DocumentReference{
+        return db.collection("playlists").document(playlistId)
+    }
+}
```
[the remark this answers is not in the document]
BartLucien: thank you

BartLucien:
> I know most of this code was already there previously, but I feel there are many inconsistencies between storing the user locally and on the cloud. I think one task for next week could be thinking about what data we want to store where, such that we avoid having useless redundancies. This will allow our UserProfileActivity to become a bit more clear and avoid having to go through the different combinations (ex. not on db/but local or the inverse) in checkUser().

Yes this is my plan for next week

nathanduchesne (approved):

sim0mo (approved):

PR #174, opened 2022-04-19
Adds the possibility to change the language of the application

Flamantrose321:
Once this pull request has been passed, it is imperative to :
- inherit the new activities from MyAppCompatActivity and no longer AppCompatActivity
- Add ALL the new strings of the application in the two strings.xml (English by default and French)

Changing the language is done as follows:
-a new string.xml associated with a language is created.
-manually translate everything in the base string.xml.
-When the user changes the language in the settings, this will change the Locale of the application
-each time activity is loaded, it will look for the current language and strings in the string.xml file corresponding to the desired language.
It is important to respect the two dots at the top of the comment.

### DIFF
--- a/app/src/androidTest/java/com/github/fribourgsdp/radio/WorkingLobbyActivityTest.kt
+++ b/app/src/androidTest/java/com/github/fribourgsdp/radio/WorkingLobbyActivityTest.kt
@@ -36,6 +36,7 @@ class WorkingLobbyActivityTest {
 
     @Before
     fun initIntent() {
+        LanguageManager(ctx).setLang("en")
         Intents.init()
     }
 


app/src/main/java/com/github/fribourgsdp/radio/SettingsActivity.kt
```diff
@@ -1,11 +1,112 @@
...
+            }
+
+            override fun onNothingSelected(parent: AdapterView<*>?) {}
+        }
+    }
+
+    private fun onItemSelectedHandler(adapterView: AdapterView<*>, view: View, position: Int, id: Long) {
+        val adapter: Adapter = adapterView.adapter
+        val language: Language = adapter.getItem(position) as Language
+        Toast.makeText(this@SettingsActivity, "Selected Employee: $language",Toast.LENGTH_SHORT).show()
```
BartLucien, line 94: I don't think Employee is what you meant.

app/src/main/res/values-fr/strings.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="utf-8"?>
```
BartLucien, line 1: As discussed through another channel. This will need to be terminated another time, maybe at the end once we have all the string that we want to translate.

app/src/main/res/values-fr/strings.xml
```diff
@@ -0,0 +1,80 @@
...
+    <string name="import_my_spotify_playlists">Import my Spotify playlists</string>
+    <string name="deletePlaylist">Delete</string>
+    <string name="editPlaylist">Edit</string>
+    <string name="sign_in_message">Sign in</string>
+    <string name="sign_out_message">Sign out</string>
+    <string name="click_on_a_song_to_delete_it">Click on a song to delete it.</string>
+    <string name="wait_for_pick_format">Waiting on %s to pick a song</string>
+    <string name="congrats">Congratulations!</string>
+    <string name="micPermissionError">The microphone permissions could not be updated.</string>
+    <string name="pref_locale">fa</string>
```
BartLucien, line 78: What does fa stand for?

app/src/main/res/values/strings.xml
```diff
@@ -74,5 +74,7 @@
     <string name="wait_for_pick_format">Waiting on %s to pick a song</string>
     <string name="congrats">Congratulations!</string>
     <string name="micPermissionError">The microphone permissions could not be updated.</string>
+    <string name="pref_locale">fa</string>
```
BartLucien, line 77: What does fa stand for?

BartLucien (commented):
Could do, in 2-3 sentences, explains the flow of what happens when you change the language, where, how, etc...? So that we have it easily available?

app/src/main/res/values-fr/strings.xml
Flamantrose321: This is a relic of an earlier version of language change. I forgot to remove it, thanks for the comment.

app/src/main/res/values/strings.xml
Flamantrose321: same as above

app/src/main/java/com/github/fribourgsdp/radio/SettingsActivity.kt
Flamantrose321: Yes, good point

app/src/main/res/values-fr/strings.xml
Flamantrose321: Yes that's right, I've actually just translated a few words to show that the system works. In the end, we'll just have to agree on the different translations.

app/src/main/java/com/github/fribourgsdp/radio/Language.kt
```diff
@@ -0,0 +1,11 @@
+package com.github.fribourgsdp.radio
+
+enum class Language(val id: Int, val code: String, val s: String) {
+    ENGLISH(0,"en", "English"),
+    FRENCH(1,"fr", "Français");
```
VictorSchneuwly, line 5: For your information, the `id` property here is useless because there exists in enums a default property called `ordinal` that already does that.

app/src/main/java/com/github/fribourgsdp/radio/LanguageManager.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    fun getLang() : String? {
+        return sharedPreferences.getString("lang", "en")
+    }
+
+    fun setLang(code :String){
+        val editor : SharedPreferences.Editor = sharedPreferences.edit()
+        editor.putString("lang", code)
+        editor.commit()
```
VictorSchneuwly, line 38: Here there is a warning on Cirrus that says you should use `apply` instead of `commit` so that it doesn't need to write the data to the main memory right away all the time. I let you consider if it is a good thing or not.

VictorSchneuwly (commented):
Overall, it could be a good idea to add more documentation to your public functions in case someone else wants to use them.

app/src/main/java/com/github/fribourgsdp/radio/LanguageManager.kt
Flamantrose321: Yes, I had seen it too. As a user will only change the language once in a while, there is no point in using apply and you can commit directly.

app/src/main/java/com/github/fribourgsdp/radio/Language.kt
Flamantrose321: ok it works, thanks for the information.

VictorSchneuwly (approved):

BartLucien (approved):

PR #175, opened 2022-04-24
Ensure users with same name are visible in lobby with mic permissions

nathanduchesne:
If merged, this PR will ensure if multiple users in a lobby have the same name, they are both displayed as well as their mic permissions.


nathanduchesne:
Manually tested with @Flamantrose321 as well, and the bug is now fixed 👍 , two users with the same name are both displayed.

Flamantrose321 (approved):
I tested it manually with Nathan and it works. For me, it's fine.

shanzhaiii (approved):
Yes this is a good idea

PR #177, opened 2022-04-25
Implement onBackPressed for all activities

nathanduchesne:
If merged this PR will do the following:

1. Override onBackPressed for all basic activities such that there is a logical ordering between activities. (except VoiceOverIpActvity, which should be gone soon)

2. Override onBackPressed for AddPlaylistActivity such that if a user quits before adding the newly modified playlist, a popup appears.

3. Override onBackPressed for LobbyActivity and GameActivity such that a popup appears to make sure the players really want to quit, and if the host quits -> the lobby/game is dropped, if a non-host user quits, he is removed from the game.


sim0mo:
Have you noticed that transitions when we press "back" are in a non uniform order ? I mean that sometimes the main menu comes from the right and sometimes from the left ? This is minor but this should be improved for the final product.

nathanduchesne:
> Have you noticed that transitions when we press "back" are in a non uniform order ? I mean that sometimes the main menu comes from the right and sometimes from the left ? This is minor but this should be improved for the final product.

You're right, I think this comes from the fact that sometimes the activity is re-created while sometimes it isn't.

nathanduchesne:
There is a known bug that when a non-host user quits the game, if the state of the game changes before the GameActivity is destroyed on that user's device, the handler will change the activity back to the game. This will be fixed next week.

sim0mo (approved):
Tested it on a real phone, and it makes the whole app smoother. Good job.

app/src/main/java/com/github/fribourgsdp/radio/QuitGameOrLobbyDialog.kt
```diff
@@ -0,0 +1,40 @@
...
+import android.view.LayoutInflater
+import android.view.View
+import android.view.ViewGroup
+import android.widget.Button
+import androidx.core.os.bundleOf
+import androidx.fragment.app.DialogFragment
+import androidx.fragment.app.setFragmentResult
+
+
+class QuitGameOrLobbyDialog(val ctx: Context, private val isHost: Boolean, private val isFromGameActivity: Boolean, private val user: User, private val db: Database, private val gameID: Long): DialogFragment() {
```
VictorSchneuwly, line 17: What's the point of giving `db: Database` and `gameID` here ?
I don't think you are still using it.

app/src/main/java/com/github/fribourgsdp/radio/PlayerGameHandler.kt
```diff
@@ -1,14 +1,17 @@
...
 import com.google.firebase.firestore.DocumentSnapshot
 
 class PlayerGameHandler(
-    private val gameID: Long,
+    val gameID: Long,
     private val view: GameView,
     db: Database = FirestoreDatabase()
 ): GameHandler(view, db), GameView.OnPickListener {
 
+    val database = db
```
VictorSchneuwly, line 13: It wasn't really intended to be exposed...

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
```diff
@@ -152,6 +154,21 @@ open class GameActivity : AppCompatActivity(), GameView {
         startActivity(intent)
     }
 
+    private fun returnToMainMenu() {
+
+        if (isHost) {
+            playerGameHandler.database.disableGame(playerGameHandler.gameID)
```
VictorSchneuwly, line 160: Here you should have simply done a function `playerGameHandler.disableGame`. That way you won't need to expose the database and the gameID.

It would simply be:
```kotlin
fun disableGame() {
    db.disableGame(gameID)
}
```

VictorSchneuwly (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/QuitGameOrLobbyDialog.kt
nathanduchesne: You are right, this was from a previous implementation.

app/src/main/java/com/github/fribourgsdp/radio/Database.kt
```diff
@@ -149,4 +149,13 @@ interface Database {
...
 
+    fun removeUserFromLobby(id: Long, user: User): Task<Void>
+
+    fun disableGame(id: Long): Task<Void>
+
+    fun disableLobby(gameID: Long): Task<Void>
+
+    fun removePlayerFromGame(gameID: Long, user: User): Task<Void>
+
+    fun makeSingerDone(gameID: Long, singerName: String): Task<Void>
```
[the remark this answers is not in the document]
nathanduchesne: Done!

app/src/main/java/com/github/fribourgsdp/radio/GameActivity.kt
nathanduchesne: Done! I also did similarly for the methods right below this which had a similar problem.

app/src/main/java/com/github/fribourgsdp/radio/PlayerGameHandler.kt
nathanduchesne: I unexposed it.

VictorSchneuwly (approved):

shanzhaiii (commented):
I think that this is great for the activities that require a specific behaviour on back button press, but I am not convinced about the implementation used for other activites. Are we not able to simply set parents in the activity manifest for the vast majority of activities? Using all these callbacks seems very convoluted and hard to maintain

PR #180, opened 2022-04-29
170 99 fix bug of not distinguishing songs with same name different artist

shanzhaiii:
If merged, this pull request will prevent the possible future error of getSong retrieving an arbitrary song of the playlist when two songs of the same name but different artists exist in the playlist


BartLucien (approved):
Seems good to me

nathanduchesne (approved):
Good that you use the redefinition of the equals function I wrote in the Song class.

PR #213, opened 2022-05-02
Bug fix - Correct popup tests

VictorSchneuwly:
Add root checks when tests use dialogs.

> When testing dialogs (popups), don't forget to do `.inRoot(isDialog())` before every `.check` or `.perform` so that the test gets the focus of the dialog right away and doesn't fail.


nathanduchesne (approved):
Thank you for fixing these tests to remove random behaviours and unwanted failing builds.

sim0mo (approved):
Good

PR #215, opened 2022-05-02
Reduce lyrics api calls

sim0mo:
If applied, this PR will drastically reduce number of API calls, as it will make a difference between songs that simply do not have lyrics, and songs for which lyrics fetching failed for some reason. This way, on game creation, it will only ask the API for the song for which an error of the second type has occurred.
It also handles more errors that can appear by using musixatch (HTTP 402, 404, 500, etc.) and thus avoids crashes.


app/src/main/java/com/github/fribourgsdp/radio/MusixmatchLyricsGetter.kt
```diff
@@ -32,147 +30,173 @@ interface LyricsGetter{
...
  */
 object MusixmatchLyricsGetter : LyricsGetter {
-        const val LYRICS_NOT_FOUND = "---No lyrics were found for this song.---"
+    const val LYRICS_NOT_FOUND = "---No lyrics were found for this song. Feel free to add yours !---"
+    const val BACKEND_ERROR_PLACEHOLDER = ""
 
-        class LyricsNotFoundException(val reason : String = LYRICS_NOT_FOUND) : Exception()
+    abstract class LyricsGetterException(val default : String) : Exception()
+    class NoLyricsFoundForThisSong : LyricsGetterException(LYRICS_NOT_FOUND)
+    class BackendError : LyricsGetterException("")
```
nathanduchesne, line 38: Here you could use the BACKEND_ERROR_PLACEHOLDER you defined above.

app/src/main/java/com/github/fribourgsdp/radio/MusixmatchLyricsGetter.kt
```diff
@@ -32,147 +30,173 @@ interface LyricsGetter{
...
+     */
+    private fun <T> checkStatus(
+        status: Int,
+        future: CompletableFuture<T>,
+        function: () -> Boolean, ) {
+        when {
+            status == 404 -> {
+                future.completeExceptionally(NoLyricsFoundForThisSong())
+            }
+            status >= 400 -> {
```
nathanduchesne, line 196: Couldn't this condition be > 400 instead? Since the case where status == 400 is dealt with above.

sim0mo: It is the error 404 that is dealt above ;) this line is to deal with 400 (Bad request, unlikely to happen), 401 (Unauthorized, i.e. when we exhaust our API calls) and 5xx (Server errors that might always happen).
NB: 404 is handled separately because it is a response that occurs on a normal request and a normal server behavior when there are no lyrics for this song. All other errors >=400 are unexpected behaviors.

app/src/main/java/com/github/fribourgsdp/radio/MusixmatchLyricsGetter.kt
sim0mo: yep !

app/src/main/java/com/github/fribourgsdp/radio/MusixmatchLyricsGetter.kt
```diff
@@ -32,147 +30,173 @@ interface LyricsGetter{
  * API Call doc : https://stackoverflow.com/questions/45219379/how-to-make-an-api-request-in-kotlin
  */
 object MusixmatchLyricsGetter : LyricsGetter {
-        const val LYRICS_NOT_FOUND = "---No lyrics were found for this song.---"
+    const val LYRICS_NOT_FOUND = "---No lyrics were found for this song. Feel free to add yours !---"
```
Flamantrose321, line 33: Why do you save the string in a constant instead of directly using R.string.(...)?

sim0mo: Because this object has no context. But this is not a problem since it is never displayed as-is but in SongFragment it takes the value from the strings.xml file. For clarity I shall rename it.

Flamantrose321 (approved):
He answered my questions. It's good for me.

PR #217, opened 2022-05-03
191 fix bug in playlistsongfragment

BartLucien:
If merged, this PR will remove the bug where a playlist stored online still had the possibility to click on the button to store the playlist online. It will also make a playlist test less magic concerning the behavior, removing a conversion set to list to then test the content of the list.


Flamantrose321 (approved):
seems good to me
this may allow the transformOnlineWorks test in playlistTest not to fail 1 out of 2 times on cirrus :)

VictorSchneuwly (approved):
Seems good to me 👍🏻

PR #233, opened 2022-05-07
232 172 bug fix settings test undefined behaviour

shanzhaiii:
Fixes the Settings unit tests which were saving to app specific file system instead of the mock file system which affected the settings for all tests run after the settings test. This can lead to undefined behaviour as test order is not guaranteed. 
Implements a nice wrapper for classes which use the file system to prevent code duplication ;)

### DIFF
--- a/app/src/androidTest/java/com/github/fribourgsdp/radio/activities/SettingsActivityTest.kt
+++ b/app/src/androidTest/java/com/github/fribourgsdp/radio/activities/SettingsActivityTest.kt
@@ -22,13 +22,14 @@ import com.github.fribourgsdp.radio.MainActivity
 import com.github.fribourgsdp.radio.R
 import com.github.fribourgsdp.radio.config.SettingsActivity
 import com.github.fribourgsdp.radio.config.language.Language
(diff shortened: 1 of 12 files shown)


shanzhaiii:
I have tried for around 4h to fix de code climate duplication, but it is not worth it. I have found a way to prevent code duplication for the save code, but I think it is quite ugly because it requires building a serializable object where we manually specify some fields which should not be serialized. For the loading code, it is simply not possible, because Json's generic function expects a rectified type and since we are calling it from the companion object, we are only able to give an object of type T:$Companion instead of T, unless there is something I am missing

VictorSchneuwly (approved):
It's a good thing this is corrected as we sometimes had a tough time understanding why some tests were failing.

BartLucien (approved):
It is  fine for me, it fixes part of the problem but I think that tests in SettingsActivity also modify the language so this test should also be modified. This can be done by someone else in another PR or you if you have time.

shanzhaiii:
> It is fine for me, it fixes part of the problem but I think that tests in SettingsActivity also modify the language so this test should also be modified. This can be done by someone else in another PR or you if you have time.

this should now be fixed

PR #234, opened 2022-05-08
188 implement flow in userprofile

BartLucien:
If merged, this PR will allow a user to choose what to with the playlists when signing in and signing out.
When signing in, the user can merge/import/dismiss the online playlist.
When signing out, the user can either keep the playlist and throw them away. It also recreate a default user.


app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
```diff
@@ -170,4 +164,101 @@ open class UserProfileActivity : MyAppCompatActivity() {
...
+
+    private fun keepPlaylist() : Task<Unit>{
+        Log.w(ContentValues.TAG, "KEEP")
+        return Tasks.forResult(null)
+        //when keeping playlist, decide whether only those saved locally should be kept or all of them
+
+    }
+    private fun deletePlaylist() : Task<Unit> {
+        user.removePlaylists(user.getPlaylists())
+        Log.w(ContentValues.TAG, "DELETE")
```
sim0mo, line 233: Do you want to keep this logging message (and previous ones) ?

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
```diff
@@ -170,4 +164,101 @@ open class UserProfileActivity : MyAppCompatActivity() {
...
+            MergeDismissImportPlaylistDialog.Choice.DISMISS_ONLINE -> dismissOnlinePlaylist()
+        }.addOnSuccessListener {
+            user.save(this)
+        }
+
+    }
+
+    private fun mergePlaylist() : Task<Unit>{
+        Log.w(ContentValues.TAG, "MERGE")
+        return db.getUser(user.id).continueWith {
```
sim0mo, line 181: What is the merging policy if two playlists happen to be have the same name ?

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
BartLucien: No I will delete them.

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
BartLucien: It does the policy implemented in the User class. As the playlists are stored in a set and two playlists are equals given the name, it does what set does with equals object. Whether they keep the old one or put the new one, I have to check

BartLucien: I have tested and it keeps the most recently added playlist. In our case, it corresponds to the playlist that comes from online. I have added a test in UserTest to demonstrate this behaviour.

sim0mo: Nice ! Better be sure of what happens underneath.

sim0mo (approved):

Flamantrose321 (approved):

PR #235, opened 2022-05-09
change the default audio route

Flamantrose321:


nathanduchesne (approved):
Thanks! This should allow us to have demos actually using the audio chat.

sim0mo (approved):

PR #236, opened 2022-05-09
Starter playlists

sim0mo:
If applied, this PR will add the functionality of already having basic playlists when we open the app for the first time. This allows user without spotify account/without time to spend creating custom playlists to play the game.
Moreover, this PR comes with 5 starter playlists :

- A _Chanson française_ playlist
- A playlist with only famous movie theme songs
- A playlist of famous pop songs from the 60's to the 2020's
- A playlist of the most iconic hits from classical music
- A playlist for true geeks containing video game songs


app/src/main/java/com/github/fribourgsdp/radio/data/StarterPlaylists.kt
```diff
@@ -0,0 +1,452 @@
+package com.github.fribourgsdp.radio.data
+
+import com.github.fribourgsdp.radio.data.Genre.*
+
+object StarterPlaylists {
```
Flamantrose321, line 5: Why did you choose to write the playlist in hard copy and not in a file?

sim0mo: Done !

shanzhaiii (changes requested):
Is it possible to not duplicate assets? I think this would be a very good thing

app/src/androidTest/assets/basicBlindTest.txt
```diff
@@ -0,0 +1,129 @@
...
+Thank You:Dido
+Wonderwall:Oasis
+All Of Me:John Legend
+Sexual Healing:Marvin Gaye
+Let Her Go:Passenger
+Still Loving You:Scorpions
+Should I Stay or Should I Go:The Clash
+Heavy Cross:Gossip
+Hallelujah:Jeff Buckley
+Ma philosophie:Amel Bent
```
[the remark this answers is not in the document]
sim0mo: Actually I have only checked with few of them, but it looks like most of them work. It would be cumbersome to check them all but I agree that they should ideally all have lyrics.

Flamantrose321 (approved):
you corrected my questions
ok for me

BartLucien (commented):
Just a question concerning the flow of these playlists. They are put into the user when creating a user and the don't have any lyrics. When do you plan to fetch the lyrics?

sim0mo:
> Just a question concerning the flow of these playlists. They are put into the user when creating a user and the don't have any lyrics. When do you plan to fetch the lyrics?

As usual either when a user opens a song or when (s)he launches a game with it. They are not included yet (unfortunately I did not spend time to fetch them all myself manually)

shanzhaiii (approved):
Perfect now. Thanks for the fix

PR #238, opened 2022-05-11
Design mockup

VictorSchneuwly:
In this PR, you can find the pdf of the first mockup design. It contains the logo and a few of the main activities to give you a general idea of what the app could look like. I also added some texts so that you can have an idea of what I was thinking while designing it.

The pdf is located [here](https://github.com/FribourgSDP/rad.io/blob/frontend/mockup-design/design/radio_mockup_design.pdf).

> Feel free to make any remarks directly in the comments of the pr.


BartLucien (approved):
I like the design, it looks good to me.

shanzhaiii (approved):
I absolutely love the design. My 2 points are that Futura is not free of use so maybe it is better to find a similar font on google fonts. My second point is I am not sure if the logo is scalable to the very small size of the the app logo on the home screen. Other than that I really love everything and can't wait to see it on the app !!

nathanduchesne (approved):
Wow, it looks amazing! It's funny because we gave you complete freedom on this and as it seems, you managed to design something that nobody can dislike!

Flamantrose321 (commented):
It's a nice design.
I would just say that the timer bar and the "guessing song" don't stand out enough.

VictorSchneuwly:
> It's a nice design. I would just say that the timer bar and the "guessing song" don't stand out enough.

I might need to make the timer bar brighter you're right. 👍🏻  And for the guess, this is just the hint, when the player will type it will be white the same way the other texts are. I will see how it is by default on android and decrease the transparency if it's not visible enough. 👍🏻

Flamantrose321:
> I might need to make the timer bar brighter you're right. 👍🏻 And for the guess, this is just the hint, when the player will type it will be white the same way the other texts are. I will see how it is by default on android and decrease the transparency if it's not visible enough

And another point, I'm not an expert on a dark mode, I'd almost say there's too much colour if I compare it to the apps I have on my phone.

sim0mo:
Good work ! You're obviously comfortable with design.

If you'll allow me to qualify, I'm not a fan of _blanding_, so I think the title "rad.io" (not the logo, which looks very nice to me) could have more personality. It could be more "funky". This could be done through the font (maybe with serifs ?), the alignment of the letters...
You could also use a color for "rad" and another for ".io".

Otherwise I find it very nice.

VictorSchneuwly:
> Good work ! You're obviously comfortable with design.
> 
> If you'll allow me to qualify, I'm not a fan of _blanding_, so I think the title "rad.io" (not the logo, which looks very nice to me) could have more personality. It could be more "funky". This could be done through the font (maybe with serifs ?), the alignment of the letters...
> You could also use a color for "rad" and another for ".io".
> 
> Otherwise I find it very nice.

I see what you mean. I need to update the font anyway so I'll think about this when I'll change it! 👍🏻

VictorSchneuwly:
> > I might need to make the timer bar brighter you're right. 👍🏻 And for the guess, this is just the hint, when the player will type it will be white the same way the other texts are. I will see how it is by default on android and decrease the transparency if it's not visible enough
> 
> And another point, I'm not an expert on a dark mode, I'd almost say there's too much colour if I compare it to the apps I have on my phone.

I figured that, since it's a game, it wasn't much of a an issue. But the other activities will look more like the menu with only the color at the top and on the buttons.

PR #241, opened 2022-05-12
Handle game error

VictorSchneuwly:
Handle the behavior of the game when a player crashed due to database issues.


app/src/androidTest/java/com/github/fribourgsdp/radio/unit/PlayerGameHandlerTest.kt
```diff
@@ -15,9 +18,11 @@ import java.util.*
 
 
 class PlayerGameHandlerTest {
+    private val ctx: Context = ApplicationProvider.getApplicationContext()
+
     private lateinit var mockSnapshot: DocumentSnapshot
 
-    private val sleepingTime = 50L
+    private val sleepingTime = 500L
```
BartLucien, line 25: Do you really need to increase the sleeping time?

app/src/main/java/com/github/fribourgsdp/radio/game/GameView.kt
```diff
@@ -63,12 +63,12 @@ interface GameView: Timer.DeadlineHandler {
...
      * Handle the end of the game.
      */
-    fun gameOver(finalScores: Map<String, Long>)
+    fun gameOver(finalScores: Map<String, Long>?, hasCrashed: Boolean = false)
 
     /**
      * Displays the given [String] in the dedicated TextView
      */
-    fun displayLyrics(lyrics : String)
+    fun updateLyrics(lyrics : String)
```
BartLucien, line 71: I think you should update the documentation

app/src/main/java/com/github/fribourgsdp/radio/game/GameView.kt
```diff
@@ -63,12 +63,12 @@ interface GameView: Timer.DeadlineHandler {
     /**
      * Handle the end of the game.
      */
-    fun gameOver(finalScores: Map<String, Long>)
+    fun gameOver(finalScores: Map<String, Long>?, hasCrashed: Boolean = false)
```
BartLucien, line 66: I think you should add the parameters in the doc.

app/src/androidTest/java/com/github/fribourgsdp/radio/unit/PlayerGameHandlerTest.kt
BartLucien: I think that even if it is for test of the same class, we can win some precious seconds by having the minimum time required to make the test run.

VictorSchneuwly: So I put it back to 50L but I will see with @nathanduchesne why one of them was increased to 1000L.

app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
```diff
@@ -127,6 +137,12 @@ open class GameActivity : AppCompatActivity(), GameView, Timer.Listener {
     }
 
     override fun displayGuessInput() {
+        // Close the lyrics popup if already open
+        lyricsPopup?.let {
```
sim0mo, line 141: Minor : I think codeclimate missed something : This block is repeated 4 times

VictorSchneuwly: Done

sim0mo (approved):
Good code overall. You fixed what concerned the lyrics popup so I'm ok with it.

PR #263, opened 2022-05-17
203 202 create song hints

Flamantrose321:


app/src/main/java/com/github/fribourgsdp/radio/game/GameView.kt
```diff
@@ -72,6 +73,9 @@ interface GameView: Timer.DeadlineHandler {
      */
     fun updateLyrics(lyrics : String)
 
+
+    fun addHint(songNameHint: SongNameHint)
+
```
nathanduchesne, line 78: Please add doc for this method (as in this whole interface).

app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
```diff
@@ -306,6 +322,29 @@ open class GameActivity : AppCompatActivity(), GameView, Timer.Listener {
...
+    fun updateHint(timeInSeconds: Int){
+        if(withHint && hintTextView.visibility == View.VISIBLE) {
+            val limit = ceil(gameDuration/ (ceil(songNameHint.length()/2.0)+1))
+            if(timeInSeconds - lastTime >= limit) {
+                songNameHint.addALetter()
+                hintTextView.text = this.songNameHint.toString()
+                lastTime = timeInSeconds
+            }
+        }
+    }
```
nathanduchesne, line 344: This algorithm makes it such that at the end of the timer, half of the letters in the song's title have been displayed. Is this maybe too much? I think we should ask the group to see if 1/3 of the total title is enough.

Flamantrose321: I think half the letters are good as they appear one after the other but if others think like you I will change.

app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
```diff
@@ -128,6 +134,10 @@ open class GameActivity : AppCompatActivity(), GameView, Timer.Listener {
             text = songName
             visibility = View.VISIBLE
         }
+
+        if(withHint) {
+            hintTextView.visibility = View.GONE
+        }
```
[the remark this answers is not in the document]
Flamantrose321: Yes, that's right. And addhint is called for every new song you guess.

app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
BartLucien: We tested (with @nathanduchesne) on scribble and at the end there were only 2 letters on small words. I think 1/2 is a bit too much, as one can guess the title just by building the words from the letters that have appeared. 1/3 or 1/4 letters seems a betters choice. Changing the value shouldn't be too difficult if we have to change it anyway.

app/src/main/java/com/github/fribourgsdp/radio/util/SongNameHint.kt
```diff
@@ -0,0 +1,53 @@
...
+
+
+    fun stripAccents(s : String): String {
+        var s2 = Normalizer.normalize(s, Normalizer.Form.NFD)
+        s2 = s2.replace("[\\p{InCombiningDiacriticalMarks}]".toRegex(), "")
+        return s2
+    }
+
+    fun replaceWithLine(s : String):String{
+        val re = Regex("[A-Za-z0-9]")
```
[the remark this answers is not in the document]
Flamantrose321: Oui c'est cela. Toutes les lettres ( a, b, c A, b c, ç, é, ö, â, .....) et chiffres sont remplacé par des _

app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
nathanduchesne: Thanks for changing it to 1/3. Everything is fine for me now.

sim0mo (approved):
Ok with this feature ! It makes the game very playable

PR #267, opened 2022-05-19
Fix bug where default playlist were disappearing when changing username

BartLucien:
If merged, this PR will remove the bug that would remove all playlist from a non google user when changing its username.

### DIFF
--- a/app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
@@ -1,24 +1,19 @@
 package com.github.fribourgsdp.radio.data.view
 
-import android.content.ContentValues
-import android.content.Context
 import android.content.Intent
 import android.graphics.PorterDuff
 import android.graphics.PorterDuffColorFilter
 import android.os.Bundle
-import android.util.Log
-import android.view.KeyEvent
 import android.view.View
-import android.view.inputmethod.InputMethodManager
 import android.widget.Button
 import android.widget.EditText
 import android.widget.ImageView
 import android.widget.TextView
 import com.github.fribourgsdp.radio.*
+import com.github.fribourgsdp.radio.auth.*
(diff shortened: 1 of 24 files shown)


app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
```diff
@@ -122,9 +122,11 @@ open class UserProfileActivity : MyAppCompatActivity(), KeepOrDismissPlaylistDia
         if(user.isGoogleUser){
             db.setUser(user.id,user)
         }else{
+            val userPlaylists = user.getPlaylists()
             val userWithoutPlaylist = user
-            userWithoutPlaylist.removePlaylists(user.getPlaylists())
+            userWithoutPlaylist.removePlaylists(userPlaylists)
             db.setUser(user.id,userWithoutPlaylist)
+            user.addPlaylists(userPlaylists)
```
VictorSchneuwly, line 129: Why remove the playlist to then add them later?

BartLucien: When an user is not signed in, only his username is saved online, not his playlists, this is why I need to remove all the  playlists from the user before saving it. I have to reput the playlist in the original user as in reality userWithoutPlaylist and  user refer to the same entity. I create the UserWithoutPlaylist just for clarity reason.

Flamantrose321, line 129: Why does it save the name in the database even if you are not connected to Google?

BartLucien: The idea was to save the name of the user on the server. After discussion with @VictorSchneuwly, it seems that we don't use it at all, as a user gives its name directly to the lobby. This behavior and uniformisation of what a offline user can do will be treated in my other task, where I have to add the offline/online behavior.

VictorSchneuwly (approved):
Seems good to me now

Flamantrose321 (approved):

PR #269, opened 2022-05-19
255 fix bug when importing playlist with spotify

BartLucien:


app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
```diff
@@ -200,7 +206,8 @@ open class UserProfileActivity : MyAppCompatActivity(), KeepOrDismissPlaylistDia
             user.name = it.result.name
             user.id = it.result.id
             user.isGoogleUser = true
-            db.setUser(user.id,user)
+            val onlineUser = user.onlineCopy()
+            db.setUser(onlineUser.id,onlineUser)
```
Flamantrose321, line 210: lines 186-190 and 206-210 are same

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockAddPlaylistActivity.kt
```diff
@@ -0,0 +1,25 @@
...
+        Mockito.`when`(db.setUser(Mockito.anyString(),any())).thenReturn(Tasks.forResult(null))
+        return db
+    }
+
+    //this is usefull in order to be able to use any() from mockito
+    private fun <T> any(): T {
+        Mockito.any<T>()
+        return uninitialized()
+    }
+    private fun <T> uninitialized(): T = null as T
```
shanzhaiii, line 24: I feel like this doesn't really have its place as a private method in the activity and should be extracted into a tools class for tests. Also, why is uninitialized() not inlined into any?

shanzhaiii (approved):
A very minor request but other than that it is a good idea and will cut down on the semi-unexpected behaviour when signing in

app/src/androidTest/java/com/github/fribourgsdp/radio/mockimplementations/MockAddPlaylistActivity.kt
BartLucien: I did both things. I also removed this initialization in the other classes I was using it.

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
BartLucien: Even though these line are the same, I personally prefer to have each button having its own behavior written directly. I think it makes more sense semantically. What do you think?

VictorSchneuwly: I see what you mean but I agree with @Flamantrose321 that this may be a better idea to make a method out of it (with `db` as a param) as those exact code snippets are also used in `PlaylistSongsFragment.kt` and `AddPlaylistActivity.kt`

Flamantrose321: yes agree with @VictorSchneuwly

BartLucien: I think I did what you asked for. Let me know if it is okay.

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
```diff
@@ -119,7 +119,10 @@ open class UserProfileActivity : MyAppCompatActivity(), KeepOrDismissPlaylistDia
         user.name = usernameField.text.toString()
         //at this point, the userId should be the firebaseUser.uid
         if(user.isGoogleUser){
-            db.setUser(user.id,user)
+            //remove all n
+            user.onlineCopyAndSave()
+            //val onlineUser = user.onlineCopy()
+            //db.setUser(onlineUser.id,onlineUser)
```
VictorSchneuwly, line 125: You forgot to remove those comments ;)

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
```diff
@@ -119,7 +119,10 @@ open class UserProfileActivity : MyAppCompatActivity(), KeepOrDismissPlaylistDia
         user.name = usernameField.text.toString()
         //at this point, the userId should be the firebaseUser.uid
         if(user.isGoogleUser){
-            db.setUser(user.id,user)
+            //remove all n
```
VictorSchneuwly, line 122: What is `n`?

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
BartLucien: You really have falcon eye, good catch.

VictorSchneuwly: Falcon gif

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
BartLucien: I don't remember but I also removed this comment

VictorSchneuwly (approved):
It's all good to me now.

PR #271, opened 2022-05-22
Fix playlist visualization not updating after google sign in

shanzhaiii:
If merged, this PR will fix the bug of the recyclerview displaying a user's playlists not updating when a playlist is deleted, as well as provide methods to easily fix the bug of playlists not updating in that same recyclerview when a user is imported

### DIFF
--- a/app/src/main/java/com/github/fribourgsdp/radio/data/view/PlaylistsFragmentHolderActivity.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/data/view/PlaylistsFragmentHolderActivity.kt
@@ -1,5 +1,6 @@
 package com.github.fribourgsdp.radio.data.view
 
+import android.content.Intent
 import android.os.Bundle
 import com.github.fribourgsdp.radio.config.MyAppCompatActivity
 import com.github.fribourgsdp.radio.util.MyFragment
@@ -9,10 +10,15 @@ class PlaylistsFragmentHolderActivity : MyAppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_playlists_fragment_holder)
-
         //initialise songs recycler view fragment
         val bundle = Bundle()
         bundle.putString(PLAYLIST_DATA, intent.getStringExtra(PLAYLIST_DATA))
         MyFragment.beginTransaction<PlaylistSongsFragment>(supportFragmentManager, bundle)
+    }
+
+    override fun onBackPressed() {
+        val intent = Intent(this, UserProfileActivity::class.java)
+        startActivity(intent)
+        finish()
     }
 }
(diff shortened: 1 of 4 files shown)


shanzhaiii:
The provided updateFragment() method in UserProfileActivity will be useful to fix the bug of playlists not visually updating when they are imported through google sign in

app/src/main/java/com/github/fribourgsdp/radio/data/view/PlaylistsFragmentHolderActivity.kt
```diff
@@ -1,5 +1,6 @@
 package com.github.fribourgsdp.radio.data.view
 
+import android.content.Intent
```
VictorSchneuwly, line 3: A minor change but I think you can remove this import.

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
```diff
@@ -114,6 +114,10 @@ open class UserProfileActivity : MyAppCompatActivity(), KeepOrDismissPlaylistDia
         userIcon = findViewById(R.id.userIcon)
     }
 
+    private fun updateFragment() {
+        val fragment = supportFragmentManager.findFragmentByTag(FRAGMENT_TAG)!! as UserPlaylistsFragment
+        fragment.notifyUserChanged()
+    }
```
VictorSchneuwly, line 120: I don't see where use assigns the tag to a `UserPlaylistsFragment`. Was the tag meant to be used on line 70?

Also, why looking for the fragment every time and not put it in a value?

Flamantrose321 (commented):
Can you also fix the fact that the name does not update properly after the sign_in?

shanzhaiii:
> Can you also fix the fact that the name does not update properly after the sign_in?

I think it should not be done in this PR because this PR adds Recyclerview syncing functionality with a big emphasis on providing ways to communicate with activities launching several levels of intents and still being able to communicate with the caller. This was a challenge and is not related to the issue for usernames.

app/src/main/java/com/github/fribourgsdp/radio/data/view/PlaylistsFragmentHolderActivity.kt
shanzhaiii: fixed in latest commit

app/src/main/java/com/github/fribourgsdp/radio/data/view/UserProfileActivity.kt
shanzhaiii: You are correct, I forgot to assign the tag during initialization. The reason we are using tags instead of putting fragment in a value is that the Fragment manager takes a fragment class as input, and takes care of building its own instance, so we do not have access to the instance.

I still think this is cleaner that inflating a recyclerview instead of using the fragmentManager to replace a RecyclerViewContainer because it leaves all the fragment lifecycle handling to be correctly handled by the fragmentManager

If you know of a cleaner way, I will gladly implement it ;)

VictorSchneuwly: OK I see why you do that now 👍🏻

VictorSchneuwly (approved):
This seems good to me now 👍🏻

Flamantrose321:
> > Can you also fix the fact that the name does not update properly after the sign_in?
> 
> I think it should not be done in this PR because this PR adds Recyclerview syncing functionality with a big emphasis on providing ways to communicate with activities launching several levels of intents and still being able to communicate with the caller. This was a challenge and is not related to the issue for usernames.

Ok if you can do this in another pr that's fine

Flamantrose321 (approved):
I tested manually and it actually fixed the bug.

PR #272, opened 2022-05-25
250 make homogeneous the access to musixmatch

BartLucien:
If merged, this PR will uniformize the requests to Musixmatch and save the fetched lyrics into FireStore.

### DIFF
--- a/app/src/androidTest/java/com/github/fribourgsdp/radio/ui/SongFragmentTest.kt
+++ b/app/src/androidTest/java/com/github/fribourgsdp/radio/ui/SongFragmentTest.kt
@@ -81,7 +81,7 @@ class SongFragmentTest {
         Espresso.onView(ViewMatchers.withId(R.id.editTextLyrics))
             .check(ViewAssertions.matches(ViewMatchers.withText(lyrics)))
     }
-    /*@Test
+    @Test
     fun getLyricsInSongFragment() {
         val bundle = Bundle()
         val user : User = User("Test User")
@@ -100,8 +100,8 @@ class SongFragmentTest {
         Espresso.onView(ViewMatchers.withId(R.id.editTextLyrics))
             .check(ViewAssertions.matches(
                 ViewMatchers.withText(MockLyricsGetter.truckfightersLyrics)))
-    }*/
-    /*
+    }
+
     @Test
     fun goodHintForSongWithoutLyrics(){
         val bundle = Bundle()
@@ -118,5 +118,5 @@ class SongFragmentTest {
         val scenario = launchFragmentInContainer<SongFragment>(bundle)
         Espresso.onView(ViewMatchers.withId(R.id.editTextLyrics))
             .check(ViewAssertions.matches(ViewMatchers.withHint(R.string.add_your_lyrics)))
-    }*/
+    }
 }
--- a/app/src/androidTest/java/com/github/fribourgsdp/radio/ui/SongFragmentTest.kt
+++ b/app/src/androidTest/java/com/github/fribourgsdp/radio/ui/SongFragmentTest.kt
@@ -81,7 +81,7 @@ class SongFragmentTest {
         Espresso.onView(ViewMatchers.withId(R.id.editTextLyrics))
             .check(ViewAssertions.matches(ViewMatchers.withText(lyrics)))
     }
-    @Test
+    /*@Test
     fun getLyricsInSongFragment() {
         val bundle = Bundle()
         val user : User = User("Test User")
@@ -100,7 +100,8 @@ class SongFragmentTest {
         Espresso.onView(ViewMatchers.withId(R.id.editTextLyrics))
             .check(ViewAssertions.matches(
                 ViewMatchers.withText(MockLyricsGetter.truckfightersLyrics)))
-    }
+    }*/
+    /*
     @Test
     fun goodHintForSongWithoutLyrics(){
         val bundle = Bundle()
@@ -117,5 +118,5 @@ class SongFragmentTest {
         val scenario = launchFragmentInContainer<SongFragment>(bundle)
         Espresso.onView(ViewMatchers.withId(R.id.editTextLyrics))
             .check(ViewAssertions.matches(ViewMatchers.withHint(R.string.add_your_lyrics)))
-    }
+    }*/
 }
--- a/app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
@@ -156,17 +156,26 @@ data class Playlist (override var name: String, var genre: Genre) : Nameable {
     }
 
     /**
-     * @return true if all songs have lyrics
+     * @return true if all songs have lyrics or have tried to fetch lyrics from Musixmatch
      */
-    fun allSongsHaveLyrics(): Boolean{
-        return songs.all { it.lyrics != "" }
+    fun allSongsHaveLyricsOrHaveTriedFetchingSome(): Boolean{
+        return songs.all { it.lyrics != MusixmatchLyricsGetter.BACKEND_ERROR_PLACEHOLDER  }
+    }
+
+    /**
+     * @return true if all song has lyrics
+     */
+    fun allSongHaveLyrics(): Boolean{
+        return songs.all { it.lyrics != MusixmatchLyricsGetter.BACKEND_ERROR_PLACEHOLDER
+                && it.lyrics != MusixmatchLyricsGetter.LYRICS_NOT_FOUND_PLACEHOLDER}
     }
 
     /**
      * @return true if no song has lyrics
      */
     fun noSongHaveLyrics(): Boolean{
-        return songs.all { it.lyrics == "" || it.lyrics == "---"}
+        return songs.all { it.lyrics == MusixmatchLyricsGetter.BACKEND_ERROR_PLACEHOLDER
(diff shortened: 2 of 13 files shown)


app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
```diff
@@ -112,6 +120,31 @@ data class Playlist (override var name: String, var genre: Genre) : Nameable {
...
+     */
+    fun loadLyrics(lyricsGetter: LyricsGetter = MusixmatchLyricsGetter) : Task<Void>{
+
+        val tasks = mutableListOf<Task<Void>>()
+        for (song in songs){
+            if(song.lyrics == "") {
+                tasks.add(
+                    Tasks.forResult(lyricsGetter.getLyrics(song.name, song.artist).thenAccept {
+                        song.lyrics = it
+                    }.join())
```
sim0mo, line 135: Does this join() make this method blocking ?

app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
```diff
@@ -120,6 +153,20 @@ data class Playlist (override var name: String, var genre: Genre) : Nameable {
...
+        return songs.all { it.lyrics != "" }
+    }
+
+    /**
+     * @return true if no song has lyrics
+     */
+    fun noSongHaveLyrics(): Boolean{
+        return songs.all { it.lyrics == "" || it.lyrics == "---"}
+    }
+
```
sim0mo, line 169: Could you please use the two constants defined in MusixmatchLyricsGetter ?

sim0mo, line 169: Why do these 2 methods not consider the same criteria of _having no lyrics_ ? it means that if every song in a playlist have lyrics = --- , it will consider that all songs in pl have lyrics ?

app/src/main/java/com/github/fribourgsdp/radio/data/view/SongFragment.kt
```diff
@@ -41,14 +41,14 @@ open class SongFragment : MyFragment(R.layout.fragment_song) {
...
-    private fun fetchLyrics(lyricsGetter: LyricsGetter) {
+    /*private fun fetchLyrics(lyricsGetter: LyricsGetter) {
         lyricsGetter.getLyrics(song.name, song.artist)
             .thenAccept{f ->
                 currentLyrics = f
                 doSaveLyrics = true
                 updateLyrics(requireView().findViewById(R.id.editTextLyrics))
             }
-    }
+    }*/
```
sim0mo, line 51: Can be removed if no longer useful

sim0mo (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/data/view/PlaylistSongsFragment.kt
```diff
@@ -81,7 +87,17 @@ open class PlaylistSongsFragment : MyFragment(R.layout.fragment_playlist_display
...
+        importLyricsButton.setOnClickListener {
+              playlist.loadLyrics().addOnSuccessListener {
+                user.addPlaylist(playlist)
+                user.save(requireContext())
+                if(playlist.savedOnline){
+                   user.onlineCopyAndSave()
+                }
+                importLyricsButton.visibility = View.INVISIBLE
+            }
+        }
```
nathanduchesne, line 100: Would it be possible to test this? I feel like clicking a button is easy coverage we can get. You may want to be careful and try mocking the LyricsGetter to avoid real calls to MusixMatch though but you can look at what was done in the spotifyTests, w.r.t how to mock JsonParser and HttpClient and it should be fine.

nathanduchesne (changes requested):

app/src/main/java/com/github/fribourgsdp/radio/data/view/AddPlaylistActivity.kt
```diff
@@ -128,23 +125,30 @@ open class AddPlaylistActivity : DatabaseHolder, MyAppCompatActivity(), SavePlay
...
+        val t = if (generateLyricsCheckBox.isChecked) {
+            playlist.loadLyrics()
+            //loadLyrics(playlist)
+        } else {
+            Tasks.forResult(null)
+        }
+        t.continueWith {
+            //playlist = it.result
+            val playlistTask = if (online) playlist.transformToOnline().addOnSuccessListener {
+                playlist.saveOnline()
```
nathanduchesne, line 140: I've just seen that saveOnline() calls db.registerSong() for every song in the playlist, which returns a Task<Void> but does not wait for the result, wouldn't this cause a problem if db.registerPlaylist(this) finishes before a song has finished being registered? (I know this is not from this weeks PR but what do you think?)

app/src/main/java/com/github/fribourgsdp/radio/data/view/PlaylistSongsFragment.kt
BartLucien: Yes, indeed. I didn't do it the way you proposed though. As load lyrics is already tested in PlaylistText, I directly mock this function.

app/src/main/java/com/github/fribourgsdp/radio/data/view/SongFragment.kt
BartLucien: Yes I will remove it.

app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
BartLucien: Yes indeed. I created these method in order to be able to test whether we should make a musixmatch call or not. In this context, having lyrics or having no lyrics from musixmatch results in the same behavior. For me, the second definition makes more sense and another name should be found for the first function. In any cases, I was planning on using the method of song that returns whether it has lyrics or not that you told me you would implement.

BartLucien: Yes of course, in any cases, I was planning on using the method of song that returns whether it has lyrics or not that you told me you would implement.

app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
BartLucien: No, as the join is withing another asynchronous object, the Task.

app/src/main/java/com/github/fribourgsdp/radio/data/view/AddPlaylistActivity.kt
BartLucien: Yes you are right, I corrected it in the Playlist.kt file.

app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
BartLucien: I created a new one so that the methods are clearer.

app/src/androidTest/java/com/github/fribourgsdp/radio/ui/SongFragmentTest.kt
```diff
@@ -117,5 +118,5 @@ class SongFragmentTest {
         val scenario = launchFragmentInContainer<SongFragment>(bundle)
         Espresso.onView(ViewMatchers.withId(R.id.editTextLyrics))
             .check(ViewAssertions.matches(ViewMatchers.withHint(R.string.add_your_lyrics)))
-    }
+    }*/
```
sim0mo, line 121: If the automatic fetching of lyrics on Song Fragment opening is still enabled, you can keep these tests I think.

app/src/main/java/com/github/fribourgsdp/radio/data/Playlist.kt
sim0mo: Oh I see the one you mean. We will se which PR will be merged first but we should look at this carefully when we will merge

sim0mo: Yeah that's better

app/src/androidTest/java/com/github/fribourgsdp/radio/ui/SongFragmentTest.kt
BartLucien: Yes indeed I will reput them.

app/src/main/java/com/github/fribourgsdp/radio/data/view/SongFragment.kt
BartLucien: As discussed over any channel of discussion, this functionality should actually stay so I will reput it.

nathanduchesne (approved):
All good for me now! I’ll let you deal with Simon’s comments now.

sim0mo:
When in the playlist edition menu, when we fetch lyrics for all songs, the app freezes for some time. If there is no way of making it fully asynchronous, a little loading indicator (even a Toast) would be appreciated. 

Otherwise it works well

BartLucien:
> When in the playlist edition menu, when we fetch lyrics for all songs, the app freezes for some time. If there is no way of making it fully asynchronous, a little loading indicator (even a Toast) would be appreciated.
> 
> Otherwise it works well

I changed the text of the button so that it is written loading lyrics when loading the lyrics.

sim0mo (approved):
You fixed/changed what was requested. Ok for me

PR #273, opened 2022-05-26
Swipe to delete song with snackbar

shanzhaiii:
If merged, this PR will replace song clicking to delete in the AddPlaylistActivity with song swiping for deletion


app/src/main/java/com/github/fribourgsdp/radio/data/view/SongSwipeHelper.kt
```diff
@@ -0,0 +1,77 @@
...
+    ): Int {
+        return makeMovementFlags(0, LEFT)
+    }
+
+    override fun onMove(
+        recyclerView: RecyclerView,
+        viewHolder: RecyclerView.ViewHolder,
+        target: RecyclerView.ViewHolder
+    ): Boolean {
+        return false
```
VictorSchneuwly, line 28: Why would `onMove` be always false?

app/src/main/res/layout/activity_add_playlist.xml
```diff
@@ -2,6 +2,7 @@
 <androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/AddPlaylistRootView"
```
VictorSchneuwly, line 5: I can't find where you use this id again. Is it used implicitly? If not, I think you can remove it as it may prevent you from having some merge conflicts. ;)

VictorSchneuwly:
Some more remarks after testing:

- I think that users are more used to ask to undo on delete rather than confirming.
- You could change the text `"Click on a song to delete it."` to `"Swipe on a song to delete it."`

app/src/main/java/com/github/fribourgsdp/radio/data/view/SongSwipeHelper.kt
shanzhaiii: I assume move is for rearranging cards in the recyclerview. This is not something we are trying to do.

app/src/main/res/layout/activity_add_playlist.xml
shanzhaiii: I am using it in an onTouchListener to cancel the deletion if the user clicks anywhere except the snackbar (unless he clicks one the the layout's elements but this is a detail)

shanzhaiii:
> Some more remarks after testing:
> 
>     * I think that users are more used to ask to undo on delete rather than confirming.
> 
>     * You could change the text `"Click on a song to delete it."` to `"Swipe on a song to delete it."`

I made the decision to have delete confirmation instead of deletion then undo because I wanted to avoid accidental deletions as much as possible. If you think the problem is the visual aspect of a whole appearing, I can have the card snap back on swipe release. If you think the problem is usability, I can change to deletion then undo functionality.

And yes I will fix the string

app/src/main/res/layout/activity_add_playlist.xml
VictorSchneuwly: > unless he clicks one the the layout's elements

I didn't see that. I don't think this is really intuitive either. The snackbar should be the only place where you could confirm (or undo) the deletion in my opinion.

VictorSchneuwly:
> > Some more remarks after testing:
> > ```
> > * I think that users are more used to ask to undo on delete rather than confirming.
> > 
> > * You could change the text `"Click on a song to delete it."` to `"Swipe on a song to delete it."`
> > ```
> 
> I made the decision to have delete confirmation instead of deletion then undo because I wanted to avoid accidental deletions as much as possible. If you think the problem is the visual aspect of a whole appearing, I can have the card snap back on swipe release. If you think the problem is usability, I can change to deletion then undo functionality.
> 
> And yes I will fix the string

If you can fix the blank space, I think it will look better 👍🏻 As to either undo or confirm we have to discuss it as a team.

nathanduchesne (commented):
After testing, I had two remarks which I would like to signal:

- When swiping to delete, I don't know why but I was expecting a red background, simply because this is what I am used to with delete on swipe interfaces. Would this be hard to do? (I have no clue if it's only changing the swiper background color or much more).
- Second of all, it's very nice that we can delete songs when creating the playlist, but it would be very cool if we could also do so when looking at our user playlists. When clicking on one of my playlists, I was expecting to be able to also delete them, but turns out it's only a feature in the playlist creation activity. I understand doing this would require more work (since deleting from playlist after creation leads to having to delete stuff from the DB), but I'm curious about what others think.

app/src/main/res/layout/activity_add_playlist.xml
shanzhaiii: It is a little bit hard to imagine just when reading the code but it makes sense if you use it. Basically the snackbar is on a timer and if you don't click delete in time it will stop the deletion process. Adding touch listener's means one can cancel the deletion process more quickly by clicking anywhere else on the screen. Try it and tell me what you think

shanzhaiii:
> After testing, I had two remarks which I would like to signal:
> 
>     * When swiping to delete, I don't know why but I was expecting a red background, simply because this is what I am used to with delete on swipe interfaces. Would this be hard to do? (I have no clue if it's only changing the swiper background color or much more).
> 
>     * Second of all, it's very nice that we can delete songs when creating the playlist, but it would be very cool if we could also do so when looking at our user playlists. When clicking on one of my playlists, I was expecting to be able to also delete them, but turns out it's only a feature in the playlist creation activity. I understand doing this would require more work (since deleting from playlist after creation leads to having to delete stuff from the DB), but I'm curious about what others think.

- I could probably have a red background but this is not necessarily very simple. I will wait and see what the group thinks about delete vs undo button
- I don't think it makes much sense to swipe to delete in the purely viewing fragment because this is exactly what the edit button in that fragment is for: it opens the playlist back into the activity which my PR is working on

nathanduchesne:
> > > After testing, I had two remarks which I would like to signal:
> > > ```
> > > * When swiping to delete, I don't know why but I was expecting a red background, simply because this is what I am used to with delete on swipe interfaces. Would this be hard to do? (I have no clue if it's only changing the swiper background color or much more).
> > > 
> > > * Second of all, it's very nice that we can delete songs when creating the playlist, but it would be very cool if we could also do so when looking at our user playlists. When clicking on one of my playlists, I was expecting to be able to also delete them, but turns out it's only a feature in the playlist creation activity. I understand doing this would require more work (since deleting from playlist after creation leads to having to delete stuff from the DB), but I'm curious about what others think.
> > > ```
> > 
> > 
> > 
> > * I could probably have a red background but this is not necessarily very simple. I will wait and see what the group thinks about delete vs undo button
> > * I don't think it makes much sense to swipe to delete in the purely viewing fragment because this is exactly what the edit button in that fragment is for: it opens the playlist back into the activity which my PR is working on
> 
> Okok for the red background, it's fine then. I personally have no problem with the "validate delete" button, the thing is it doesn't work on my phone, meaning if I press anywhere on the screen (I've tried everywhere literally), it still deletes the song, meaning the "protection against unwanted deletions" doesn't work in this case.
> 
> Now concerning the comment I made on the playlist viewing, I didn't see the edit button so I was a bit frustrated I could only delete it. Forget what I said about this.

After playing around with it a bit more, turns out it doesn't always happen, sometimes the delete works and sometimes it doesn't..

shanzhaiii:
The other alternative was preferred by the group

PR #275, opened 2022-05-27
Swipe to delete with undo

shanzhaiii:
This is a variant of #273 where instead of Song deletion requiring a confirmation in the snackbar, songs are directly deleted by default when swiped but can be restored with the snackbar if it was a mistake.
I leave it up to the team to approve its favorite means.


nathanduchesne (approved):
Just tested it and it works fine, without the bug we had in the other PR, all good for me!

Flamantrose321 (approved):
i have tested both features and this is the one i prefere

PR #276, opened 2022-05-27
Fix Google Sign In Playliststs not updating

shanzhaiii:
If merged, this will fix the bug of playlists not updating after a sign in.
For some reason, updating the recyclerview fragment is not sufficient so we manually refresh the activity, which is ok since sign in is so uncommon. (same method as when changing languages)


BartLucien:
The signing in procedure is nice, just 1 thing. I think the same should be done when signing out.

BartLucien (approved):
Seems good to me

PR #281, opened 2022-06-01
Try to fix multiple games bug

Flamantrose321:
If applied, this PR will fix the major bug of not being able to play more than one game or rad.io

### DIFF
--- a/app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
+++ b/app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
@@ -167,10 +167,7 @@ open class GameActivity : AppCompatActivity(), GameView, Timer.Listener {
         showLyricsButton.visibility = View.GONE
 
         // Show the edit text and the submit button instead
-        songGuessEditText.apply {
-            text.clear()
-            visibility = View.VISIBLE
-        }
+        songGuessEditText.visibility = View.VISIBLE
         songGuessSubmitButton.visibility = View.VISIBLE
     }
 
@@ -281,6 +278,10 @@ open class GameActivity : AppCompatActivity(), GameView, Timer.Listener {
             }
             false
         }
+
+        // clear edittext on click
+        songGuessEditText.setOnClickListener { songGuessEditText.text.clear() }
+
         muteButton = findViewById(R.id.muteChannelButton)
         muteButton.setOnClickListener {
             voiceChannel.mute()


app/src/main/java/com/github/fribourgsdp/radio/database/FirestoreDatabase.kt
```diff
@@ -560,11 +567,28 @@ class FirestoreDatabase(var refMake: FirestoreRef, var transactionMgr: Transacti
...
+
+    override fun removeLobbyListener(){
+        metadataGameListerRegistration?.remove()
+    }
+    override fun removeMetadataGameListener(){
+        lobbyListerRegistration?.remove()
+    }
+
+    override fun removePublicLobbyListener() {
+        publicLobbyListerRegistration?.remove()
```
VictorSchneuwly, line 586: Here I would also put the attributes back to `null`.

app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
```diff
@@ -68,6 +68,7 @@ open class GameActivity : AppCompatActivity(), GameView, Timer.Listener {
     protected lateinit var voiceChannel: VoiceIpEngineDecorator
 
     private lateinit var playerGameHandler: PlayerGameHandler
+    private lateinit var hostGameHandler : HostGameHandler
```
VictorSchneuwly, line 71: Instead of doing a late init here, since not everyone has it, I would make it nullable.
Then in `onDestroy()`, line 112, you can simply do a null check: `hostGameHandler?.unlinkFromDatabase()`. I think this is a better implementation since some user won't have some `lateinit` not initialized.

VictorSchneuwly (commented):
I now this is still a draft but here are some comments I had on your code.

app/src/main/java/com/github/fribourgsdp/radio/game/GameActivity.kt
Flamantrose321: corrected

app/src/androidTest/java/com/github/fribourgsdp/radio/activities/WorkingJoinGameActivityTest.kt
```diff
@@ -93,6 +99,23 @@ class WorkingJoinGameActivityTest {
     }
 
 
+    @Test
+    fun pressBackWorks(){
+        val context: Context = ApplicationProvider.getApplicationContext()
+
+        val intent = Intent(context, VoiceOverIPActivity::class.java)
+        ActivityScenario.launch<VoiceOverIPActivity>(intent).use { _ ->
```
VictorSchneuwly, line 107: Why are you launching `VoiceOverIPActivity` in the tests for the `JoinActivity`?

app/src/main/java/com/github/fribourgsdp/radio/voip/MyIRtcEngineEventHandler.kt
```diff
@@ -12,8 +12,9 @@ class MyIRtcEngineEventHandler (private val appCompatActivity: AppCompatActivity
         super.onActiveSpeaker(uid)
         val activeSpeakerView : TextView = appCompatActivity.findViewById(R.id.activeSpeakerView)
         activeSpeakerView.invalidate()
-        activeSpeakerView.text = mapHashToName[uid]?.let {
+        val s = mapHashToName[uid]?.let {
             appCompatActivity.getString(R.string.active_speaker_format, it)
         } ?: ""
+        activeSpeakerView.text = s
```
VictorSchneuwly, line 18: It wasn't working before?

app/src/androidTest/java/com/github/fribourgsdp/radio/activities/WorkingJoinGameActivityTest.kt
Flamantrose321: well seen

app/src/main/java/com/github/fribourgsdp/radio/voip/MyIRtcEngineEventHandler.kt
Flamantrose321: It should also work without

app/src/androidTest/java/com/github/fribourgsdp/radio/activities/WorkingJoinGameActivityTest.kt
BartLucien: If that is the case, also use the rule so that the same activity is not launched twice.

VictorSchneuwly (approved):
Good to have this fixed

PR #282, opened 2022-06-01
Fix serialization issues in chanson francaise playlist

sim0mo:
PR to solve serialization bug regarding Chanson Française playlist.
Reformat : use constant instead of hardcoded string


Flamantrose321 (approved):
Perfect 
Thanks Mauron for finding the bug

PR #290, opened 2022-06-02
Correction EditText reset between rounds

VictorSchneuwly:
The click listener on pr #285 was actually deleting too much. I changed again the clear but this time I set the text to an empty string. 

You may ask:
> But why not use `text.clear()` ?

For some undetermined reason, it had the behavior to sometime make the `EditText` disappear. That's why I opted for the empty string.


BartLucien (approved):
Tested behavior is good

nathanduchesne:
> The click listener on pr #285 was actually deleting too much. I changed again the clear but this time I set the text to an empty string.
> 
> You may ask:
> 
> > But why not use `text.clear()` ?
> 
> For some undetermined reason, it had the behavior to sometime make the `EditText` disappear. That's why I opted for the empty string.

I was actually asking myself this exact question. Thanks!

nathanduchesne (approved):

PR #291, opened 2022-06-04
Tests Cleanup and Firestore string modularization

sim0mo:
Cleans all test classes by :
- Extracting strings and values
- Changing vars to vals where possible
- Fixing typos
- Removing lambda declarations where possible
- Optimizing imports
- Replacing `assert` by `assertTrue`
- Using existing methods where possible (`getAndCast`)

All top-level constants are now located in single file TestConstants.kt

Also extracts all strings related to Firestore database and puts them in new file called FirestoreConstants.kt


app/src/androidTest/java/com/github/fribourgsdp/radio/utils/TestConstants.kt
```diff
@@ -0,0 +1,38 @@
+package com.github.fribourgsdp.radio.utils
```
BartLucien, line 1: I think this class could be put in the test folder

nathanduchesne (approved):
Great work, thanks!

BartLucien (approved):
Good job.

